PR #2, opened 2023-04-05
Tray: Add more buttons

TheKodeToad:
debatable whether they are needed, I think they would be useful

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -28,7 +28,7 @@ function initTray(win: BrowserWindow) {
             click: createAboutWindow
         },
         {
-            label: "Force Update Vencord",
+            label: "Update Vencord",
             async click() {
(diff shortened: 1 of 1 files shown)


Vendicated:
this is a good idea and in fact I actually also planned adding them here!

maybe also add "Update Vencord" from the menu

TheKodeToad:
I didn't add the Force Update option as that seems like something you wouldn't need often.

Vendicated:
that option is for when Vencord breaks as a fallback so I think it'd be good to have down there, Discord does too

Vendicated:
that option is for when Vencord breaks as a fallback so I think it'd be good to have down there, Discord does too

TheKodeToad:
Done now!

Vendicated:
thank you!!

PR #7, opened 2023-04-09
Tray: Option to hide

TheKodeToad:
Another trivial change!
Having a tray icon can often feel cluttered ;p
if the user doesn't switch off minimise to tray now... uhh... I'm wondering what the best solution for this is :P

### DIFF
--- a/src/renderer/components/Settings.tsx
+++ b/src/renderer/components/Settings.tsx
@@ -23,6 +23,12 @@ export default function SettingsUi() {
 
     const switches: [keyof typeof Settings, string, string, boolean?][] = [
         [
+            "tray",
+            "Use tray",
+            "Add a tray icon for Vencord Desktop",
+            true
+        ],
+        [
             "minimizeToTray",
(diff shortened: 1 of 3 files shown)


Vendicated:
Whenever you got time, please rebase and you can now use settings change listeners to show/hide tray without retsrat ( i pinged u on discord with an example!! )

TheKodeToad:
Should be ready! :P

src/renderer/components/Settings.tsx
```diff
@@ -21,12 +21,19 @@ export default function SettingsUi() {
         Button
     } = Common;
 
-    const switches: [keyof typeof Settings, string, string, boolean?][] = [
+    const switches: [keyof typeof Settings, string, string, boolean?, (() => boolean)?][] = [
+        [
+            "tray",
+            "Use tray",
```
Vendicated, line 27:
```suggestion
            "Tray Icon",
```

PR #17, opened 2023-04-19
Static title

zt64:
Changes the main window to use "Vencord" as its title. It's much nicer than having 200 character long window title that changes every-time you navigate


Vendicated:
hop on settings toggle

zt64:
@Vendicated review NOW

PR #32, opened 2023-05-14
Add back transparency as a settings option

KidOfCubes:


Vendicated:
Hi! This is already handled by Vencord itself so there is no need to add it here. thank you anyway

KidOfCubes:
I can't seem to get it working on the latest build though, could you maybe take a look?
I've added the transparency config values I have on my vencord installation into  `/home/kidofcubes/.config/VencordDesktop/VencordDesktop/settings.json` and `/home/kidofcubes/.config/VencordDesktop/VencordDesktop/settings/settings.json`, and it still has a white background

PR #46, opened 2023-07-05
feat: implement transparency for Windows

### LINKED ISSUE #16: [Feature Request] Acrylic Support
While this is a **Windows** only solution, it does add some very nice additions for theme creators.

And since it's not exactly night and day to simply switch out an ASAR for transparency anymore, I think if Vencord Desktop added support for Vibe it would attract more users.

More about vibe here: https://github.com/pykeio/vibe

Yumiro:
introduces window transparency for Windows via Electron's new API. the API is only supported for Windows 11 22H2 and above.

todo:
- [x] ~~provide a basic transparent theme for users~~ add a disclaimer for needing a transparent theme
- [x] add a way to detect if the user's Windows version is 11 22H2 or above? if not just add a disclaimer

closes #16


Vendicated:
>  add a way to detect if the user's Windows version is 11 22H2 or above

```js
import { release } from "os";

const supportsStuff = process.platform === "win32" && Number(release().split(".").pop()) >= 22621;
```

Yumiro:
Should work now, let me know if there's a better approach since I'm new to all of this

Vendicated:
no, you did it perfectly! thank you

src/renderer/components/Settings.tsx
```diff
@@ -97,6 +98,45 @@ export default function SettingsUi() {
                 </Switch>
             ))}
 
+            {supportsWindowsTransparency && (
+                <>
+                    <Switch
+                        value={Settings.transparent ?? false}
```
Vendicated, line 104: wouldn't it make more sense to completely remove this switch and instead merge it into the below select? So that the default state of the select is No transparency and then you can choose one of the options

Yumiro: done, i've also added back the full restart disclaimer and the theme disclaimer. as for where people would find them, i don't know either - i'm only aware of the [theme](https://github.com/uwu/Acrylic/blob/v2/app/theme.css) that Acrylic uses as well as any "transparent" themes on BetterDiscord that might more-or-less work if edited properly ([like this one](https://[https://github.com/MeowIce/TranslucencePlus]))

Vendicated (approved):
I think this is good to merge, unless you have anything else you would like to change?

Yumiro:
Sounds good to me, I don't think there's anything else needed to change

Vendicated:
thank you!!

PR #47, opened 2023-07-08
feat: use inset title bar style on macOS

ryanccn:
<table>
<tr>
 <td>Normal Discord</td>
 <td>Vencord Desktop</td>
 <td>This PR</td>
</tr>
<tr>
 <td></td>
 <td></td>
 <td></td>
</tr>
</table>

### DIFF
--- a/src/renderer/patches/platformClass.tsx
+++ b/src/renderer/patches/platformClass.tsx
@@ -1,0 +1,30 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0
+ * Vencord Desktop, a desktop app aiming to give you a snappier Discord Experience
+ * Copyright (c) 2023 Vendicated and Vencord contributors
+ */
+
+import { addPatch } from "./shared";
+
+addPatch({
+    patches: [
+        {
+            find: "platform-web",
+            replacement: {
+                // eslint-disable-next-line no-useless-escape
+                match: /return (?=__OVERLAY__\?""\.concat\((\i))/,
+                replace: "$1=$self.getPlatformClass(); return "
+            }
+        }
+    ],
+
+    getPlatformClass() {
+        const platform = navigator.platform.toLowerCase();
+
+        if (platform.includes("mac")) return "platform-osx";
+        if (platform.includes("win")) return "platform-win";
+        if (platform.includes("linux")) return "platform-linux";
+
+        return "platform-web";
+    }
+});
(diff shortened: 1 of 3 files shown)


src/main/ipc.ts
```diff
@@ -41,6 +41,10 @@ ipcMain.on(IpcEvents.GET_VERSION, e => {
     e.returnValue = app.getVersion();
 });
 
+ipcMain.on(IpcEvents.GET_PLATFORM, e => {
+    e.returnValue = process.platform;
```
Vendicated, line 45: this doesn't seem necessary. In preload you can just access it directly, in rendered you can use navigator.platform

src/renderer/macos.css
```diff
@@ -0,0 +1,49 @@
+html[data-vcd-platform="macos"] .wrapper-1_HaEi {
```
Vendicated, line 1: is having our own styles even necessary? Discord already has their own platform class they add to the html element, which is for example `platform-win` on windows

Vendicated: image

Vendicated: yeah, the platform-osx class has almost the same styles as you attached

src/renderer/patches/platformClass.tsx
```diff
@@ -0,0 +1,30 @@
...
+                replace: "$1=$self.getPlatformClass(); return "
+            }
+        }
+    ],
+
+    getPlatformClass() {
+        const platform = navigator.platform.toLowerCase();
+
+        if (platform.includes("mac")) return "platform-osx";
+        if (platform.includes("win")) return "platform-win";
```
Vendicated, line 25: does this not cause issues? Since now it expects there to be the entire Discord titlebar

Vendicated: Yes, platform-win leads to there being no padding at the top 
image

platform-linux on the other hand has no effect. As such I would remove the windows and linux logic and only use the osx class

src/renderer/patches/platformClass.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { addPatch } from "./shared";
+
+addPatch({
+    patches: [
+        {
+            find: "platform-web",
+            replacement: {
+                // eslint-disable-next-line no-useless-escape
+                match: /return (?=__OVERLAY__\?""\.concat\((\i))/,
+                replace: "$1=$self.getPlatformClass(); return "
```
Vendicated, line 16: i don't like that this patch assigns to the variable. If Discord ever updates their bundler and this becomes a const, it will likely lead to a crash.

i suggest something like
```js
match: /(?<=" platform-overlay"\):)\i/,
replace: "$self.getPlatformClass()"
```

Vendicated (approved):

PR #48, opened 2023-07-08
feat: use standardized icon for macOS

ryanccn:
Designed using Apple's official Sketch design template.


Vendicated:
what's the difference?

ryanccn:
@Vendicated it uses the rounded square icon format that has been the standard from Big Sur onwards

Vendicated:
please show a screenshot

ryanccn:
CleanShot 2023-07-09 at 17 11 32@2x

Vendicated:
This code doesn't seem right. You removed the setDock line and MACOS_ICON_PATH seems unused

ryanccn:
Yeah that constant can probably be removed, you don’t need a `setIcon` call on macOS

Vendicated:
i think the static icon shouldnt be needed anymore?

i also think this icon would also apply to other platforms, so maybe rename the file to build/icon.macos.icns and specify in package.json electron-builder config

ryanccn:
> i think the static icon shouldnt be needed anymore?

yes, I'll remove that

> i also think this icon would also apply to other platforms, so maybe rename the file to build/icon.macos.icns and specify in package.json electron-builder config

it wouldn't, icns is a macOS-specific format

Vendicated:
> i also think this icon would also apply to other platform

image

Vendicated:
WRONG BUTTON

Vendicated:
thank u btw

PR #50, opened 2023-07-09
feat: add hide menu bar items on macOS

ryanccn:


src/main/mainWindow.ts
```diff
@@ -133,6 +133,18 @@ function initMenuBar(win: BrowserWindow) {
                         app.quit();
                     }
                 },
+                ...(process.platform === "darwin"
```
Vendicated, line 136: i'm not a fan of these ternary array spreads. It would be more readable to have something like so:
```ts
[
	process.platform === "darwin" && {
		label: "Hide",
		role: "hide"
	}
].filter(Boolean);
```

Vendicated: see latest commit

PR #51, opened 2023-07-09
fix: improve app hiding functionality

ryanccn:
- Always uses hiding functionality on macOS since the tray doesn't exist and it's the behavior of mainline Discord
- Use `app.hide()` instead of `win.hide()` for built-in activation behavior

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -274,10 +274,17 @@ function createMainWindow() {
     win.setMenuBarVisibility(false);
 
     win.on("close", e => {
-        if (isQuitting || Settings.store.minimizeToTray === false || Settings.store.tray === false) return;
+        if (
+            isQuitting ||
+            (process.platform !== "darwin" &&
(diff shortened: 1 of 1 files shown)


src/main/mainWindow.ts
```diff
@@ -274,10 +274,17 @@ function createMainWindow() {
     win.setMenuBarVisibility(false);
 
     win.on("close", e => {
-        if (isQuitting || Settings.store.minimizeToTray === false || Settings.store.tray === false) return;
+        if (
+            isQuitting ||
+            (process.platform !== "darwin" &&
+                (Settings.store.minimizeToTray === false || Settings.store.tray === false))
```
Vendicated, line 280: this boolean should be extracted into a variable to make the code more readable

```js
const useTray = Settings.store.minimizeToTray !== false && Settings.store.tray !== false;
if (isQuitting || (!useTray && process.platform !== "darwin")) return;
```

Vendicated (approved):
looks good thank you

PR #52, opened 2023-07-09
feat: add splash window theming

ryanccn:
Obtains the `--text-normal` and `--background-primary` CSS variables from the CSSOM after document load, saves them in Vencord Desktop settings, and overrides `--fg` and `--bg` in the splash screen with these values if they exist.

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -11,6 +11,7 @@ import {
     dialog,
     Menu,
     MenuItemConstructorOptions,
+    nativeTheme,
     Tray
 } from "electron";
 import { rm } from "fs/promises";
@@ -341,7 +342,9 @@ function createMainWindow() {
     removeSettingsListeners();
     removeVencordSettingsListeners();
 
-    const { staticTitle, transparencyOption, enableMenu, discordWindowsTitleBar } = Settings.store;
+    const { staticTitle, transparencyOption, splashTheming, splashBackground, enableMenu, discordWindowsTitleBar } =
+        Settings.store;
+
     const { frameless, macosTranslucency } = VencordSettings.store;
 
     const noFrame = frameless === true || (process.platform === "win32" && discordWindowsTitleBar === true);
@@ -361,7 +364,7 @@ function createMainWindow() {
         ...(transparencyOption && transparencyOption !== "none"
             ? {
                   backgroundColor: "#00000000",
-                  backgroundMaterial: Settings.store.transparencyOption,
+                  backgroundMaterial: transparencyOption,
                   transparent: true
               }
             : {}),
@@ -371,7 +374,14 @@ function createMainWindow() {
                   vibrancy: "sidebar",
                   backgroundColor: "#ffffff00"
               }
-            : {}),
+            : {
+                  backgroundColor: splashTheming
+                      ? splashBackground
+                      : nativeTheme.shouldUseDarkColors
+                      ? "#313338"
+                      : "#ffffff",
+                  transparent: false
+              }),
         ...(process.platform === "darwin" ? { titleBarStyle: "hiddenInset" } : {}),
         ...getWindowBoundsOptions(),
         autoHideMenuBar: enableMenu
--- a/src/main/splash.ts
+++ b/src/main/splash.ts
@@ -9,6 +9,8 @@ import { join } from "path";
 import { SplashProps } from "shared/browserWinProperties";
 import { ICON_PATH, VIEW_DIR } from "shared/paths";
 
+import { Settings } from "./settings";
+
 export function createSplashWindow() {
     const splash = new BrowserWindow({
         ...SplashProps,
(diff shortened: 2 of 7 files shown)


ryanccn:
Huh, types are fine in VS Code

src/main/mainWindow.ts
```diff
@@ -274,17 +282,22 @@ function createMainWindow() {
...
+                  backgroundColor: Settings.store.splashTheming
+                      ? Settings.store.splashBackground
+                      : nativeTheme.shouldUseDarkColors
+                      ? "#313338"
+                      : "#ffffff",
                   transparent: false
               }),
         ...(Settings.store.staticTitle ? { title: "Vencord" } : {}),
-        ...(VencordSettings.store.macosTranslucency
+        ...((VencordSettings.store.macosTranslucency as boolean)
```
Vendicated, line 300: what is this cast for?

src/main/splash.ts
```diff
@@ -9,10 +9,19 @@ import { join } from "path";
 import { SplashProps } from "shared/browserWinProperties";
 import { VIEW_DIR } from "shared/paths";
 
-export function createSplashWindow() {
+import { Settings } from "./settings";
+
+export async function createSplashWindow() {
```
Vendicated, line 14: there doesn't seem to be the need for this to be async
```suggestion
export function createSplashWindow() {
```

Vendicated:
> Huh, types are fine in VS Code

That likely means our typescript dependency is outdated

src/renderer/themedSplash.ts
```diff
@@ -0,0 +1,26 @@
+/*
+ * SPDX-License-Identifier: GPL-3.0
+ * Vencord Desktop, a desktop app aiming to give you a snappier Discord Experience
+ * Copyright (c) 2023 Vendicated and Vencord contributors
+ */
+
+import { Settings } from "./settings";
+
+window.addEventListener("load", () => {
```
Vendicated, line 9: Technically this shouldn't be necessary, but i would prefer if you checked document.readyState first to make sure this callback always runs even if the document is somehow already loaded

Much similar to here https://github.com/Vencord/Vesktop/blob/main/src/preload/index.ts#L29-L35

Vendicated: Also, as mentioned on Discord, with this approach the variables won't be properly updated if you change theme in any way. One solution would be to use the beforeunload event to save them another time right before closing the app. Though I'm not sure if this is actually called if you Alt+F4 (or equivalent). If no, there are likely similar electron apis that can achieve this

Vendicated (changes requested):

Vendicated:
hmm, it seems like this doesn't work with Discord's nitro themes. It would be really nice if they were also supported

Vendicated:
fg-semi-trans should also be changed so the border has the proper colour

src/main/splash.ts
```diff
@@ -8,11 +8,28 @@ import { BrowserWindow } from "electron";
...
 export function createSplashWindow() {
     const splash = new BrowserWindow(SplashProps);
 
     splash.loadFile(join(VIEW_DIR, "splash.html"));
 
+    const { splashBackground, splashColor, splashTheming } = Settings.store;
+    if (splashTheming) {
+        if (splashColor) {
+            splash.webContents.insertCSS(`body { --fg: ${splashColor} !important }`);
+            const semiTrans = tinycolor(splashColor);
```
Vendicated, line 24: this adds a ~30kb library for such a small thing

It's also not necessary. You can obtain parsed css colours via getComputedStyle. For example for background-primary, `getComputedStyle(document.querySelector("[class|=chatContent]")).backgroundColor` gives you it in rgba format

ryanccn: It seems that all the classes that use `--text-normal` are hashed, and you can't really rely on the elements that use these classes to be available on load anyways

package.json
```diff
@@ -29,6 +29,7 @@
         "@fal-works/esbuild-plugin-global-externals": "^2.1.2",
         "@types/node": "^18.15.11",
         "@types/react": "^18.0.33",
+        "@types/tinycolor2": "^1.4.3",
```
Vendicated, line 32:
```suggestion
```

Vendicated:
hmm i tested (with clearvision) and it doesn't seem to do anything for me

src/main/utils/color.ts
```diff
@@ -0,0 +1,42 @@
...
+        if (match) {
+            const [, str] = match;
+            return `#${str}${Math.round(alpha * 255)
+                .toString(16)
+                .padStart(2, "0")}`;
+        }
+    }
+
+    return original;
+};
```
Vendicated, line 42: this is still absolute insanity

someone proposed this a while back, why not use it?

```js
function resolveColor(color: string) {
    const span = document.createElement("span")
    span.style.color = color;
    
    document.body.append(span)
    const rgbColor = getComputedStyle(span).color;
    span.remove();
    
    return rgbColor;
}

ryanccn: adopted!

Vendicated (approved):

Vendicated:
cool, thanks!

Vendicated:
oh no..........

PR #53, opened 2023-07-09
feat: add "Reset Vesktop" option to menu & tray

ryanccn:


ryanccn:
Not sure whether to make this button also clear Vencord Desktop settings or not, could be a good idea

Vendicated:
How do you even use this entry now that we hid the menubar?

ryanccn:
Should I also add this in Vencord Desktop Settings then?

(It's available on macOS since the menu bar is always there)

Vendicated:
having it in the tray would likely be the most logical

image

src/main/mainWindow.ts
```diff
@@ -104,6 +119,29 @@ function initTray(win: BrowserWindow) {
     });
 }
 
+async function clearData(win: BrowserWindow) {
+    const { response } = await dialog.showMessageBox(win, {
+        message: "Are you sure you want to reset Vencord Desktop?",
+        detail: "Vencord Desktop will automatically restart after this operation.",
+        buttons: ["Yes", "No"],
+        cancelId: 1,
+        defaultId: 0,
```
Vendicated, line 128: consider using a const enum for this

src/main/mainWindow.ts
```diff
@@ -104,6 +119,29 @@ function initTray(win: BrowserWindow) {
     });
 }
 
+async function clearData(win: BrowserWindow) {
+    const { response } = await dialog.showMessageBox(win, {
+        message: "Are you sure you want to reset Vencord Desktop?",
```
Vendicated, line 124:
```suggestion
        message: "Are you sure you want to reset Vencord Desktop? This will log you out and clear other local settings.",
```

Vendicated (changes requested):
Aside my other proposed changes,
- Please update the name from Vencord Desktop to Vesktop
- "reset Vesktop" kind of implies that Vesktop settings are also reset, which is not actually the case. We should either also reset those settings or rephrase it

Vendicated (approved):

Vendicated:
thank you!

PR #54, opened 2023-07-10
feat: enable secure keyboard entry on password fields

ryanccn:


ryanccn:
Chromium does this by default

PR #85, opened 2023-08-05
feat: portable zip

lewisakura:
some people don't want to use the nsis-based portable executable to run vesktop, so give them the option of using a zip-based version


Vendicated:
i don't like having two different portable builds. removing the portable exe is probably better

PR #103, opened 2023-08-28
ci: switch to vedantmgoyal2009/winget-releaser

lewisakura:
Works on Linux which should in theory be faster than a Windows runner, + it uses Komac which ironically has much better support for Winget manifest generation than Microsoft's own tooling.

### DIFF
--- a/.github/workflows/winget-submission.yml
+++ b/.github/workflows/winget-submission.yml
@@ -1,33 +1,17 @@
-# Based on Microsoft/DevHome Winget CI, modified for use in Vencord/Vesktop.
-#
-# Copyright (c) Microsoft Corporation and Contributors
-# Licensed under the MIT license.
-
 name: Submit to Winget Community Repo
 
 on:
-  workflow_dispatch:
   release:
     types: [published]
 
 jobs:
   winget:
     name: Publish winget package
-    runs-on: windows-latest
-    env:
-      WINGET_PAT: ${{ secrets.WINGET_PAT }}
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/winget-submission.yml
```diff
@@ -1,33 +1,17 @@
...
-          # Fetching latest release from GitHub
-          $github = Invoke-RestMethod -uri "https://api.github.com/repos/vencord/vesktop/releases"
-          $targetRelease = $github | Select-Object -First 1
-          $installerUrl = $targetRelease | Select-Object -ExpandProperty assets -First 1 | Where-Object -Property name -match 'Vesktop-Setup.*?exe' | Select-Object -ExpandProperty browser_download_url
-          $packageVersion = $targetRelease.tag_name.Trim("v")
-
-          # Update package using wingetcreate
-          Invoke-WebRequest https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
-          .\wingetcreate.exe update $packageId --version $packageVersion --urls "$installerUrl" --submit --token $env:WINGET_PAT
+        uses: vedantmgoyal2009/winget-releaser@v2
```
Vendicated, line 13: never use tags for untrusted actions, always pin to long commit hash

Vendicated: especially in cases like this where you're passing tokens to it

exhq:
:trollface:

Vendicated:
is this mergeable? i have no idea

mantikafasi:
Try and see

lewisakura:
yeah it should be fine

PR #108, opened 2023-08-31
fix: add correct WMClass to desktop file

### LINKED ISSUE #96: Unable to pin Vesktop due to mismatched WMClass on Linux
Running 0.3.0 out of the box. Vesktop loads perfectly fine, icon loads, except I can't right click to pin Vesktop to my dock. This action works fine for every other app I have except Vesktop.

Digging into `/usr/share/applications/vencorddesktop.desktop` I see that:
```
StartupWMClass=Vesktop
```
However, running `xprop WM_CLASS` and selecting Vesktop gives me:
```> xprop WM_CLASS
WM_CLASS(STRING) = "vencorddesktop", "VencordDesktop"
```

Setting `StartupWMClass` to "vencorddesktop" and running `update-desktop-database` fixes the issue of being unable to pin Vesktop. However then the icon breaks (at least on my system, with no particular obvious reason as to why), however I fixed this just by manually setting `Icon` to `/usr/share/icons/hicolor/0x0/apps/vencorddesktop.png`.

Assuming that this is reproducible, it should just be a simple update to the .desktop file.

lewisakura:
It seems the WMClass on Linux still uses the old internal VencordDesktop name, so this PR adds the correct WMClass to the desktop file to resolve it.

Resolves #96.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,8 @@
                 "GenericName": "Internet Messenger",
                 "Type": "Application",
                 "Categories": "Network;InstantMessaging;Chat;",
-                "Keywords": "discord;vencord;electron;chat;"
+                "Keywords": "discord;vencord;electron;chat;",
+                "WMClass": "VencordDesktop"
             }
         },
         "mac": {


Vendicated:
I assume this is because we still use the name Vencorddesktop as name property. changing it would mean that everyone loses their settings which is meeh, maybe we gotta figure out a solution to that and properly migrate

lewisakura:
> I assume this is because we still use the name Vencorddesktop as name property. changing it would mean that everyone loses their settings which is meeh, maybe we gotta figure out a solution to that and properly migrate

Yeah this is why I didn't just blanket rename the package, but this should be a quick fix for Linux users regardless when 0.3.1 comes out.

PR #120, opened 2023-09-22
Optimize the menu bar for macOS

X1nto:


src/main/mainWindow.ts
```diff
@@ -182,13 +182,31 @@ function initMenuBar(win: BrowserWindow) {
             }
         },
         isDarwin && {
```
Vendicated, line 184: its way less ugly if u do it like this
```js
...(!isDarwin ? [] : [
	all the stuff here
])
```

src/main/mainWindow.ts
```diff
@@ -181,14 +183,34 @@ function initMenuBar(win: BrowserWindow) {
...
-        },
+        ...(!isDarwin ? [] : [
+            {
+                type: "separator"
+            },
+            {
+                label: "Settings",
+                accelerator: "CmdOrCtrl+,",
+                async click() {
+                    mainWin.webContents.executeJavaScript("Vencord.Webpack.Common.SettingsRouter.open()")
```
Vendicated, line 194: did you even test this? this gives you a blank screen
```suggestion
                    mainWin.webContents.executeJavaScript("Vencord.Webpack.Common.SettingsRouter.open('My Account')")
```

X1nto: Worked fine for me, weird

X1nto: Just realized this seems to open whatever the last page in the settings was, so no wonder it displays a blank screen on the first try.

Vendicated:
lint is failing

PR #121, opened 2023-09-22
MacOs: properly request microphone and camera permission

X1nto:


src/main/index.ts
```diff
@@ -50,6 +50,14 @@ function init() {
 
         registerScreenShareHandler();
         bootstrap();
+        
+        //TODO we should only ask these permissions when the user joins a voice chat/video call for the first time
+        //TODO we should also handle the granted result respectively. How does this official client handle permissions?
+        systemPreferences.askForMediaAccess('microphone')
```
Vendicated, line 56: you can most likely call these from within electron permission handler

X1nto: you mean https://www.electronjs.org/docs/latest/api/session#sessetdevicepermissionhandlerhandler?

src/main/mediaPermissions.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+import { session, systemPreferences } from "electron";
+
+export function registerMediaPermissionsHandler() {
+    if (process.platform !== "darwin") return;
+
+    session.defaultSession.setPermissionRequestHandler((_webContents, permission, callback, details) => {
+        if (permission === "media") {
+            if (details.mediaTypes!.includes("audio")) {
```
Vendicated, line 14: no reason not to just use safer optional chaining
```suggestion
            if (details.mediaTypes?.includes("audio")) {
```

package.json
```diff
@@ -94,7 +94,13 @@
                     ]
                 }
             ],
-            "category": "Network"
+            "category": "Network",
+            "extendInfo": {
+                "NSMicrophoneUsageDescription": "This app needs access to the microphone",
+                "NSCameraUsageDescription": "This app needs access to the camera",
```
Vendicated, line 100: would be better to give them more descriptive descriptions

X1nto: I copied them from Discord. Should I go with the previous version (before 1953e4e07807597ffa9a7db653741b087881561c)?

Vendicated: discord moment 😭

X1nto: Didn't really answer my question 😛

src/main/mediaPermissions.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function registerMediaPermissionsHandler() {
+    if (process.platform !== "darwin") return;
+
+    session.defaultSession.setPermissionRequestHandler((_webContents, permission, callback, details) => {
+        if (permission === "media") {
+            if (details.mediaTypes?.includes("audio")) {
+                systemPreferences.askForMediaAccess("microphone").then(callback);
+            }
+            if (details.mediaTypes?.includes("video")) {
```
Vendicated, line 17:
```suggestion
            else if (details.mediaTypes?.includes("video")) {
```

src/main/mediaPermissions.ts
```diff
@@ -0,0 +1,22 @@
...
+                systemPreferences.askForMediaAccess("camera").then(callback);
+            }
```
Vendicated, line 19:
```suggestion
            } else callback(true);
```

src/main/mediaPermissions.ts
Vendicated: actually, as pointed out by lumap, it might be both at once. in any case, the default case should call callback(true)

X1nto: Uh, my current implementation supports both at once. It's not if-else, it's if and if. Default case being callback(true) makes no sense to me, could you elaborate?

src/main/mediaPermissions.ts
X1nto: Default case being callback(true) makes no sense to me, could you elaborate?

src/main/mediaPermissions.ts
X1nto: Hm, I guess that makes sense.

X1nto: As for the if blocks, your suggestion of using `else if` instead of `if` makes the permission requests mutually exclusive. My initial implementation makes sure that both can be requested at the same time.

src/main/mediaPermissions.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright (c) 2023 Vendicated and Vencord contributors
+ */
+
+import { session, systemPreferences } from "electron";
+
+export function registerMediaPermissionsHandler() {
+    if (process.platform !== "darwin") return;
+
+    session.defaultSession.setPermissionRequestHandler(async (_webContents, permission, callback, details) => {
+        let granted: boolean = true;
```
Vendicated, line 13:
```suggestion
        let granted = true;
```

PR #127, opened 2023-09-27
added the beginnings of screensharing with audio support (requires virtmic binary)

kaitlynkittyy:
The binary is from [here](https://github.com/edisionnano/Screenshare-with-audio-on-Discord-with-Linux)
Direct [Download](https://github.com/edisionnano/Screenshare-with-audio-on-Discord-with-Linux/blob/main/virtmic?raw=true)

(Binary works only on pipewire, instructions for pulseaudio in the linked repo)
Following the instructions from the screenshare repo, you can select an application in the binary and then stream it.


src/renderer/patches/screenShareAudio.ts
```diff
@@ -0,0 +1,39 @@
+const original = navigator.mediaDevices.getDisplayMedia;
```
Vendicated, line 1: make sure this only runs on the target platform and doesnt error if virtmic isnt active

kaitlynkittyy: Should work now i believe?

PR #128, opened 2023-09-27
fix: appimage not relaunching

zyrouge:
Fixes an issue where AppImage does not relaunch.

> Relevant issue: https://github.com/electron-userland/electron-builder/issues/1727.
> Fix taken from [revoltchat/desktop](https://github.com/revoltchat/desktop/blob/2373e7d6887b57ef8b837d6be53e2fa8b24a6fe0/src/main.ts#L309).


Vendicated:
thanks

PR #131, opened 2023-09-28
fix audio and quality settings menu on wayland

kaitlynkittyy:
Still a work in progress, but its definitely moving along


src/main/screenShare.ts
```diff
@@ -23,17 +26,22 @@ export function registerScreenShareHandler() {
...
+                    ? {
+                          width: 1920,
+                          height: 1080
+                      }
+                    : {
+                          width: 176,
+                          height: 99
+                      }
+            })
+            .catch(() => null);
```
Vendicated, line 42: you should log this error

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -98,7 +98,7 @@ export function openScreenSharePicker(screens: Source[], skipPicker = false) {
                 }
             }
         );
-    });
+    }).catch(null);
```
Vendicated, line 101:
```suggestion
    });
```

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -71,7 +71,7 @@ addPatch({
     }
 });
 
-export function openScreenSharePicker(screens: Source[], skipPicker = false) {
+export async function openScreenSharePicker(screens: Source[], skipPicker: boolean) {
```
Vendicated, line 74:
```suggestion
export function openScreenSharePicker(screens: Source[], skipPicker = false) {
```

src/main/screenShare.ts
```diff
@@ -43,17 +51,21 @@ export function registerScreenShareHandler() {
...
+                if (stream === null) return callback({});
+            }
 
             callback(video ? { video: sources[0] } : {});
             return;
         }
 
         const choice = await request.frame
-            .executeJavaScript(`Vesktop.Components.ScreenShare.openScreenSharePicker(${JSON.stringify(data)})`)
+            .executeJavaScript(`Vesktop.Components.ScreenShare.openScreenSharePicker(${JSON.stringify(data)}, false)`)
```
Vendicated, line 68:
```suggestion
            .executeJavaScript(`Vesktop.Components.ScreenShare.openScreenSharePicker(${JSON.stringify(data)})`)
```

src/main/screenShare.ts
```diff
@@ -63,6 +75,7 @@ export function registerScreenShareHandler() {
         if (!choice) return callback({});
 
         const source = sources.find(s => s.id === choice.id);
+
```
Vendicated, line 78:
```suggestion
```

Vendicated (changes requested):

Vendicated:
you introduced so many unrelated changes - please don't

Vendicated:
even with these changes, it still requests twice for me. once initially when requesting the screen, then when you click ok and the stream is about to start. but that seems to be entirely out of our control

PR #151, opened 2023-10-21
Build arm64 Linux packages

FieryFlames:


Vendicated:
is there really no better way to do this less verbose??

FieryFlames:
Afaik no, I tried one target with the inner target as a array of strings but builder just errored, it's only allowed to be a string

PR #155, opened 2023-10-22
feat: allow disabling updates check

ryanccn:


src/renderer/components/Settings.tsx
```diff
@@ -47,7 +47,8 @@ export default function SettingsUi() {
             "openLinksWithElectron",
             "Open Links in app (experimental)",
             "Opens links in a new Vesktop window instead of your web browser"
-        ]
+        ],
+        ["checkUpdates", "Check for updates", "Automatically check for updates", true]
```
Vendicated, line 51:
```suggestion
        ["checkUpdates", "Check for updates", "Automatically check for Vesktop updates", true]
```

Vendicated:
thanks

PR #156, opened 2023-10-22
feat: make updater window adapt to splash colors

ryanccn:
A few things to consider:

- Should the splash colors setting be renamed if the updater also adapts to the theme?
- ~~Should the success/danger/link colors also be extracted from Discord web?~~ Implemented

### DIFF
--- a/static/views/updater.html
+++ b/static/views/updater.html
@@ -25,7 +25,7 @@
         button {
             cursor: pointer;
             padding: 0.5em;
-            color: var(--fg);
+            color: var(--bg);
             border: none;
             border-radius: 3px;
             font-weight: bold;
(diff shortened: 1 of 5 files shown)


static/views/updater.html
```diff
@@ -25,7 +25,7 @@
         button {
             cursor: pointer;
             padding: 0.5em;
-            color: var(--fg);
+            color: var(--bg);
```
Vendicated, line 28: huh?

ryanccn: This looks better on some themes and doesn't on others. Unfortunately, there isn't really a way to find the actual Discord button styles without hardcoding class names and injecting fake HTML, so maybe keep the buttons as-is?

Vendicated:
updater is gone 🦀 🦀 🦀 🦀 🦀 🦀

PR #157, opened 2023-10-22
Use local shiggy gif for splash

wearrrrr:
shiggy.gif is now a local file inside static instead of making a web request, not sure if this saves time but if something ever happens to the emoji for some reason it won't bork things
updated the name field in package.json from VencordDesktop to vencord-desktop, because package.json is really weird about capitalization in the name field.
appBadge.ts - changed `if (count === -1)` to `if (count < 0)`, this should make sure if the count was -2 (somehow) then it would still work fine

of course feel free to just close this pr or ask for any changes, I did this because I was incredibly bored and have enjoyed using vesktop :)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "VencordDesktop",
+    "name": "vencord-desktop",
```
Vendicated, line 2: this breaks everyone's settings

wearrrrr: oops, completely forgot that settings relied on that, let me fix that rq

PR #189, opened 2023-10-27
Make linux display badges

patyhank:
Add `desktopName` to package.json to make badge visible
https://github.com/signalapp/Signal-Desktop/issues/3387#issuecomment-500109546

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "homepage": "https://vencord.dev/",
     "license": "GPL-3.0",
     "author": "Vendicated <vendicated@riseup.net>",
+    "desktopName": "vencorddesktop.desktop",
     "main": "dist/js/main.js",
     "scripts": {
         "build": "tsx scripts/build/build.mts",


D3SOX:
Not sure if this is always correct because the AUR package uses `vencord.desktop`

nyakowint:
I dont think the aur package is official

D3SOX:
Hm, it's mentioned in the README. I also don't get why it even includes its own desktop file. Isn't there a pacman target for electron-builder that could be used to bring it more in line with the other packages?

uku3lig:
> I also don't get why it even includes its own desktop file.

aur maintainers are built different

jokes aside, it uses `pnpm package:dir`, which doesn't output a `.desktop` file

patyhank:
On Linux platform, only supports Unity desktop environment, you need to specify the *.desktop file name to desktopName field in package.json. By default, it will assume {app.name}.desktop.
And this only seems to affect Linux.

Vendicated:
this does nothing for me

PR #195, opened 2023-10-28
feat: allow patching screenshare with any audio device

ryanccn:
This is useful when loopback audio drivers or applications, such as BlackHole, are already installed.


ryanccn:
@Vendicated I need some help here, currently using a video device errors with `TypeError: Video was requested, but no video stream was provided`.  Testing seems to indicate that patching isn't successful? The overridden `navigator.mediaDevices.getDisplayMedia` never gets called.

pparmy:
i couldn't find a discord to ask this, but can you get banned for streaming in 1080p 60fps without nitro? would it be better for non-nitro havers to just stick  with 720p 30fps or could it slide . . ?

RinLovesYou:
Behavior on linux is rather unpredictable. 

## Starting a stream using camera

Initial result is this error, and no stream starting.
image

Afterwards, pressing the "Go Live" button will immediately start the stream using the previously selected camera. Window picker does not appear.

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -106,14 +115,24 @@ export function openScreenSharePicker(screens: Source[], skipPicker: boolean) {
...
+                                    await VesktopNative.virtmic.start([v.audioSource], v.workaround);
+                                }
                             }
+
+                            patchDisplayMedia({
+                                audioId: v.audioDevice,
+                                venmic: !!v.audioSource && v.audioSource !== "None",
+                                videoId: v.cameraId
+                            });
+
```
OpaqueReptile, line 133: Just switched over to Linux/Vesktop and was looking for a solution like this! 

I had to move patchDisplayMedia() outside of the above if-else block to get this to work. Still fails initially, but the streaming button afterwards will grab the then grab the camera every time.

CerulanLumina:
Doesn't allow picking video patch on wayland - the modal is entirely skipped since the selection is done with the desktop portal instead of in the app.

Vendicated:
a lot has changed about the code since and now it has global state that it applies inside a getUserMedia patch. that'll likely solve your issues of only applying once! could you please rebase based on that?

t1stm:
sorry for my useless comment, but this will make screen-sharing games on X11 finally bearable using my [small tool](https://github.com/t1stm/nvfbc-v4l2). 

if the dev of this pr cannot currently / doesn't have the time to rebase it, would it be possible for me to do it?

Novantric:
+1 to bump, this really should be a feature

ryanccn:
I rewrote the implementation from scratch based on the latest code and added back audio device patching support; video device patching is on the roadmap as well :>

t1stm:
i have been dailying this PR for three days now and I haven't encountered any issues yet. maybe a better UI can be implemented later, but for now i think it's OK. (huge thanks for the dev since this is what i personally needed for streaming in discord)

update 4 Aug 2025:
this PR is still being used as my daily client

ryanccn:
@Vendicated any updates on the status of this PR? Would be great to have this merged in some form before the two-year mark rolls around.

nyabinary:
Branch has conflicts also you might want to join the Discord, might have better luck reaching the devs there.

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -364,6 +372,16 @@ function StreamSettingsUi({
...
+    const [audioDevices, , audioDevicesPending] = useAwaiter(
+        () => navigator.mediaDevices.enumerateDevices().then(g => g.filter(d => d.kind === "audioinput")),
+        { fallbackValue: [] }
+    );
+
+    const [videoDevices, , videoDevicesPending] = useAwaiter(
+        () => navigator.mediaDevices.enumerateDevices().then(g => g.filter(d => d.kind === "videoinput")),
+        { fallbackValue: [] }
+    );
+
```
Vendicated, line 384:
```suggestion
    const [[audioDevices, videoDevices], _deviceError, devicesPending] = useAwaiter(
        () => navigator.mediaDevices.enumerateDevices().then(devices => [
        	devices.filter(d => d.kind === "audioinput"),
        	devices.filter(d => d.kind === "videoinput")
        ] as const),
        { fallbackValue: [[], []] as const }
    );
```

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -426,6 +444,47 @@ function StreamSettingsUi({
...
+                                    label,
+                                    value: deviceId
+                                }))}
+                                isSelected={d => settings.overrideAudioDevice === d}
+                                select={d => {
+                                    setSettings(v => ({ ...v, overrideAudioDevice: d }));
+                                }}
+                                serialize={String}
+                                popoutPosition="top"
+                                closeOnSelect={true}
```
Vendicated, line 463:
```suggestion
                                closeOnSelect
```

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -426,6 +444,47 @@ function StreamSettingsUi({
...
+                                    label,
+                                    value: deviceId
+                                }))}
+                                isSelected={d => settings.overrideVideoDevice === d}
+                                select={d => {
+                                    setSettings(v => ({ ...v, overrideVideoDevice: d }));
+                                }}
+                                serialize={String}
+                                popoutPosition="top"
+                                closeOnSelect={true}
```
Vendicated, line 483:
```suggestion
                                closeOnSelect
```

src/renderer/patches/screenShareFixes.ts
```diff
@@ -11,21 +11,70 @@ import { isLinux } from "renderer/utils";
...
+        return audioDevice?.deviceId;
+    } catch (error) {
+        return null;
+    }
+}
+
+navigator.mediaDevices.getDisplayMedia = async function (opts) {
+    const stream = await original.call(this, opts);
+
+    if (overrideDevices.audio) {
```
Vendicated, line 40: this entire block is strange. why get temporary video & audio tracks just to remove them again and call getUserMedia for a second time? you can just set deviceId and similar properties on the options object before calling the original

PR #223, opened 2023-11-08
Add showOpenDialog method to VesktopNative

c0re100:
Vesktop does not provide `showOpenDialog` method like `DiscordNative.fileManager.showOpenDialog`.
If some plugin need to open file dialog to select a directory for some tasks, then there's no way to do it.

Example: [vc-message-logger-enhanced](https://github.com/Syncxv/vc-message-logger-enhanced/blob/master/components/settings/ImageCacheSetting.tsx#L63)

So we need to provide `showOpenDialog` to a plugin developer.

To without headache, I decided to use `showOpenDialog` same as DiscordNative.
Then we can use it like this.
```
const native = IS_WEB ? DiscordNative : VesktopNative;
const [path] = await native.fileManager.showOpenDialog({ properties: ["openDirectory", "single"] });
```


Vendicated:
there's really no good argument for having this ipc method

plugins having any use for this ipc method implies that the plugin code has security issues (the renderer being able to write arbitrary files => sandbox escape), as is the case for MessageLoggerEnhanced

Instead of allowing arbitrary file write to the renderer, plugins should add ipc methods for scoped file write, and also an ipc method for showing the open dialog (handled entirely in node, so the renderer cannot manipulate its value (once again, sandbox escape))

as such, i will reject this PR. thank you regardless!

Issue #220, opened 2023-11-07
Crashes/annoying popups when Vesktop settings file is read-only


pluiedev:
As a NixOS user, I'm accustomed to writing config files via home-manager, which are installed as symlinks to a read-only file in the Nix store. Unfortunately, Vesktop doesn't seem to like that... Sometimes it crashes, sometimes it works just fine except when it tries to write to the config, like as shown here.

![image](https://github.com/Vencord/Vesktop/assets/22406910/180aaaaa-b772-41c2-8fa3-ba16eade363d)

I'm not that convinced that Vesktop *needs* to write to the config that often — is there any way to patch this?

Vendicated:
nixos is not supported.

Vendicated:
almost all software needs to persist data to the disk. im not sure what you're expecting. just don't make it readonly

Scrumplex:
> almost all software needs to persist data to the disk. im not sure what you're expecting. just don't make it readonly

Almost all software cleanly separates user settings from application data though. Obviously you are going to need a place to cache stuff, but a settings.json shouldn't really change much beyond what the user does themselves.

Many people like to manage dotfiles in a central repo and one way of preventing drift is to make those dotfiles read-only. I don't see how NixOS is even a factor, as it's just one of many systems that work that way.

Vendicated:
> Almost all software cleanly separates user settings from application data though

yes?
![grafik](https://github.com/Vencord/Vesktop/assets/45497981/7c4e9a8b-600a-4d8d-902c-a668bc6c8ddc)

i don't understand what you're trying to say, sorry

Scrumplex:
> i don't understand what you're trying to say, sorry

Judging by the [interface definition of `Settings`](https://github.com/Vencord/Vesktop/blob/main/src/shared/settings.d.ts) there are several values that aren't really user settings like windows state or `firstLaunch` and `skippedUpdate`. Obviously these are gone trigger the save handler, without the user intending to.

Those would belong in separate state files preferably in `.local/share`

I would be happy to implement this and some other improvements for read-only configs myself, if you don't really want to work on it

Vendicated:
> preferably in .local/share

no. they should be in .config, alongside the other files. putting them in a separate file is not a bad idea though. for example state.json, in the same directory as settings.json

Vendicated:
prs welcome, just make sure to preserve backwards compatibility, for example by migrating them from the old file

n3oney:
[...]

PR #224, opened 2023-11-08
refactor: introduce separate state settings

Scrumplex:
Related: #220

# TODOs
- [ ] Only take actual props when migrating from legacy settings

### DIFF
--- a/src/shared/settings.d.ts
+++ b/src/shared/settings.d.ts
@@ -18,19 +18,22 @@ export interface Settings {
     arRPC?: boolean;
     appBadge?: boolean;
     discordWindowsTitleBar?: boolean;
-
-    maximized?: boolean;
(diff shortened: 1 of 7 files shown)


src/shared/settings.d.ts
```diff
@@ -6,7 +6,8 @@
 
 import type { Rectangle } from "electron";
 
-export interface Settings {
+export interface SettingsV1 {
```
Vendicated, line 9: why introduce this versioning scheme? i don't like it

just add handling for the case that reading state.json fails with ENOENT and transfer the state data from settings.json

Vendicated:
superseded by https://github.com/Vencord/Vesktop/commit/4074e8d6ac9c322d81b93e16fabde521e1a4c0c1

thanks a lot regardless!

PR #239, opened 2023-11-16
feat: Implement MiddleClickAutoScroll setting (#171)

### LINKED ISSUE #171: Linux: Option to enable MiddleClickAutoscroll
Due to X using middle click as a keybind to paste from clipboard, Chromium disables middle click scrolling by default. However, this functionality can be re-enabled using `--enable-blink-features=MiddleClickAutoscroll` if a user so desires (as is my case when using [XMousePasteBlock](https://github.com/milaq/XMousePasteBlock)). Would be great to have an in-app option to configure this instead of having to apply it using desktop files or via the command line.

Thanks,
pointy

Nullmatic:
Resolves #171

### DIFF
--- a/src/renderer/components/Settings.tsx
+++ b/src/renderer/components/Settings.tsx
@@ -34,7 +34,7 @@ export default function SettingsUi() {
             true,
             () => Settings.tray ?? true
         ],
-        !isMac && ["middleClickScroll", "Middle Click Scroll", "Middle clicking scrolls instead of pastes", false],
+        !isMac && ["middleClickAutoscroll", "Middle Click Autoscroll", "Middle-clicking scrolls instead of pasting content from PRIMARY (Requires a full restart)", false],
         ["arRPC", "Rich Presence", "Enables Rich Presence via arRPC", false],
         [
             "disableMinSize",
--- a/src/renderer/components/Settings.tsx
+++ b/src/renderer/components/Settings.tsx
@@ -34,6 +34,7 @@ export default function SettingsUi() {
             true,
             () => Settings.tray ?? true
         ],
+        !isMac && ["middleClickScroll", "Middle Click Scroll", "Middle clicking scrolls instead of pastes", false],
         ["arRPC", "Rich Presence", "Enables Rich Presence via arRPC", false],
         [
             "disableMinSize",
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -390,7 +390,7 @@ function createMainWindow() {
             devTools: true,
             preload: join(__dirname, "preload.js"),
             spellcheck: true,
-            ...(Settings.store.middleClickScroll && {enableBlinkFeatures: 'MiddleClickAutoscroll'}) 
+            ...(Settings.store.middleClickAutoscroll && {enableBlinkFeatures: 'MiddleClickAutoscroll'}) 
         },
         icon: ICON_PATH,
         frame: !noFrame,
@@ -406,8 +406,6 @@ function createMainWindow() {
         autoHideMenuBar: enableMenu
     }));
     win.setMenuBarVisibility(false);
-
-    if(Settings.store.middleClickScroll) 
 
     win.on("close", e => {
         const useTray = !isDeckGameMode && Settings.store.minimizeToTray !== false && Settings.store.tray !== false;
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -389,7 +389,8 @@ function createMainWindow() {
             contextIsolation: true,
             devTools: true,
             preload: join(__dirname, "preload.js"),
-            spellcheck: true
+            spellcheck: true,
+            ...(Settings.store.middleClickScroll && {enableBlinkFeatures: 'MiddleClickAutoscroll'}) 
         },
         icon: ICON_PATH,
         frame: !noFrame,
@@ -405,6 +406,8 @@ function createMainWindow() {
         autoHideMenuBar: enableMenu
     }));
     win.setMenuBarVisibility(false);
+
+    if(Settings.store.middleClickScroll) 
 
     win.on("close", e => {
         const useTray = !isDeckGameMode && Settings.store.minimizeToTray !== false && Settings.store.tray !== false;
--- a/src/shared/settings.d.ts
+++ b/src/shared/settings.d.ts
@@ -12,7 +12,7 @@ export interface Settings {
     transparencyOption?: "none" | "mica" | "tabbed" | "acrylic";
     tray?: boolean;
     minimizeToTray?: boolean;
-    middleClickScroll?: boolean;
+    middleClickAutoscroll?: boolean;
     openLinksWithElectron?: boolean;
     staticTitle?: boolean;
     enableMenu?: boolean;
--- a/src/shared/settings.d.ts
+++ b/src/shared/settings.d.ts
@@ -12,6 +12,7 @@ export interface Settings {
     transparencyOption?: "none" | "mica" | "tabbed" | "acrylic";
     tray?: boolean;
     minimizeToTray?: boolean;
+    middleClickScroll?: boolean;
     openLinksWithElectron?: boolean;
     staticTitle?: boolean;
     enableMenu?: boolean;


D3SOX:
Works fine for me on KDE Plasma (X11) but the text in the clipboard is pasted anyway. Not sure if this can be prevented by Vesktop though.

https://github.com/Vencord/Vesktop/assets/24937357/76d28938-c82d-492c-981f-36e062c52b44

ZirixCZ:
With smooth scroll and without
https://github.com/Vencord/Vesktop/assets/49836430/1809e73c-60cc-4656-b666-4ec1cd94bf32

src/renderer/components/Settings.tsx
```diff
@@ -34,6 +34,7 @@ export default function SettingsUi() {
             true,
             () => Settings.tray ?? true
         ],
+        !isMac && ["middleClickAutoscroll", "Middle Click Autoscroll", "Middle-clicking scrolls instead of pasting content from PRIMARY (Requires a full restart)", false],
```
Vendicated, line 37:
```suggestion
        isLinux && ["middleClickAutoscroll", "Middle Click Autoscroll", "Middle-clicking scrolls instead of pasting content from PRIMARY (Requires a full restart)", false],
```

Vendicated: this message is also misleading. it still pastes anyway

Nullmatic: Implemented the fixes you suggested and updated the description to be less misleading

Nullmatic:
I left it out of the description since I thought it would be too long, but I wonder if I should also put that it still pastes text?

alchzh:
@D3SOX @Nullmatic can you test the plugin in https://github.com/Vendicated/Vencord/pull/2024 to test if it gets rid of the unwanted pastes?

Nullmatic:
> @D3SOX @Nullmatic can you test the plugin in [Vendicated/Vencord#2024](https://github.com/Vendicated/Vencord/pull/2024) to test if it gets rid of the unwanted pastes?

Sorry for the late reply. The plugin did limit the number of pastes when enabled on my side at least. It doesn't stop them but makes it harder for it to happen every click so that is a plus.

The funny thing I also found when testing is that when enabling native Wayland capabilities using the flag: ```--ozone-platform-hint=auto``` and having middle click scroll enabled, after the first scroll it crashes the client.

alchzh:
> Sorry for the late reply. The plugin did limit the number of pastes when enabled on my side at least. It doesn't stop them but makes it harder for it to happen every click so that is a plus.

Can you try increasing the reenable delay slider in the plug-in settings? Also check the other setting for when to allow pastes- the text input is selected until you  click on another element and by default I still allow pastes when text boxes are selected.


> The funny thing I also found when testing is that when enabling native Wayland capabilities using the flag: ```--ozone-platform-hint=auto``` and having middle click scroll enabled, after the first scroll it crashes the client. 

Should be reported to electron?

Nullmatic:
> Can you try increasing the reenable delay slider in the plug-in settings? Also check the other setting for when to allow pastes- the text input is selected until you click on another element and by default I still allow pastes when text boxes are selected.

For me, when I set the settings to "Whenever a textbox is active" and set reenable delay to 1000ms usually after the first middle-click, the second one will paste into the chat box. The same happens when setting to "Only when clicking on a textbox." I might just be clicking a bit fast for it to register perhaps? As said before though, it works and isn't instantaneous.

> Should be reported to electron?

Does this only affect electron applications or is this more of a problem with the blink engine?

D3SOX:
When this requires a Vencord plugin anyway I don't see the need for this anymore as https://github.com/Vendicated/Vencord/pull/2003 exists

D3SOX:
I can definitely see your point here now. I haven't ever tried Vencord in a standalone Chromium browser with that flag.

alchzh:
> The funny thing I also found when testing is that when enabling native Wayland capabilities using the flag: `--ozone-platform-hint=auto` and having middle click scroll enabled, after the first scroll it crashes the client.

I was unable to replicate this on Plasma 5, middle click scroll works as normal when launched with native Wayland.

Haxorzz:
> Still works well on Plasma 6, running both as an xwayland and a wayland client, and no crashes. Though it does attempt to paste whatever is on my clipboard when I autoscroll. Is it possible to prevent that? I know BetterDiscord as a modification to the desktop client disables middle-click pasting, but I have no idea if that's possible in Vesktop which I understand is not the desktop client.

There might be an electron blink launch flag to disable middle click pasting, or you can just disable it system-wide (or best effort) in plasma by going to settings -> General behavior (under workspace0 -> Middle click: Paste selected text (untick)

happenslol:
Is this still being considered? This is a very important feature for me.

I'd also note that middle-click pasting is still happening even though I have disabled it at OS-level. Webcord and all my other applications respect this, Vesktop is the only one still pasting.

Haxorzz:
> Is this still being considered? This is a very important feature for me.
> 
> I'd also note that middle-click pasting is still happening even though I have disabled it at OS-level. Webcord and all my other applications respect this, Vesktop is the only one still pasting.

for now you can add  --enable-blink-features=MiddleClickAutoscroll in the launch options, and it works

happenslol:
> for now you can add --enable-blink-features=MiddleClickAutoscroll in the launch options, and it works

For anyone coming across this - this works and also disables middle click pasting. Thanks!

Haxorzz:
> > For anyone coming across this - this works and also disables middle click pasting. Thanks!
> 
> Which desktop environment do you use? On Plasma, it still pastes when I'm using autoscrolling, but the launch option does work.

on plasma you need to disable middle-click paste in settings > under workspace - general behavior > Middle click: Pastes selected text
do note this is wayland only

happenslol:
For the record, I'm on hyprland and I have the middle click paste option disabled. Even with the option on hyprland disabled though, it would still paste. After using the flag suggested above, it now no longer pastes, but only autoscrolls.

Webcord and the regular desktop client do not have this behavior (they autoscroll and don't paste by default).

alchzh:
@nixRidge @Haxorzz @happenslol can you try my plugin in https://github.com/Vendicated/Vencord/pull/2024 ? not sure how it interacts with that launch flag

alchzh:
> @alchzh Yes, I'd love to, but sorry to ask how do I load this plugin in Vesktop? I understand that Vencord needs to be built with the plugin, but I wasn't able to figure out where to put the relevant index.ts in order to build it with the Vesktop application.

Build Vencord with the plugin then go to Vesktop Settings in your app and change it to the `Vencord/dist` directory you just built in

image

Haxorzz:
> > > Is this still being considered? This is a very important feature for me.
> > > I'd also note that middle-click pasting is still happening even though I have disabled it at OS-level. Webcord and all my other applications respect this, Vesktop is the only one still pasting.
> > 
> > 
> > for now you can add --enable-blink-features=MiddleClickAutoscroll in the launch options, and it works
> 
> on gnome, i'm still having it pasting even with middle click paste disabled via dconf settings and and this flag enabled. have been scouring around for a chromium solution to disabling middle click paste, as i'm getting mixed info on whether on wayland middle click paste is implemented by the system or the apps, with many saying it's the apps implementing the feature regardless of system settings (note firefox has a flag to disable this). hope a plugin/solution can be looked at again.

It's both. There's a system level toggle on Wayland that won't force this behavior like x11 does (which seems broken on gnome, surprise surprise). And there's per-app implementation of autoscroll (and middle click paste), so any app can technically implement it or not implement auto scrolling if they wish

LetsBlend:
@happenslol or anyone else willing to help, could you explain how you managed to disable the middle click pasting on hyprland? I am running vesktop with these flags: 
`--enable-features=UseOzonePlatform --ozone-platform=wayland --enable-blink-features=MiddleClickAutoscroll`
and i have set `middle_click_paste = false` inside the misc section in my hyprland config. Yet, vesktop still pastes selected text into my input field.

happenslol:
> @happenslol or anyone else willing to help, could you explain how you managed to disable the middle click pasting on hyprland? I am running vesktop with these flags: `--enable-features=UseOzonePlatform --ozone-platform=wayland --enable-blink-features=MiddleClickAutoscroll` and i have set `middle_click_paste = false` inside the misc section in my hyprland config. Yet, vesktop still pastes selected text into my input field.

It seems pretty much impossible to disable this for electron apps, as they implement the behavior themselves independent from the compositor. In firefox you can disable it in `about:config`, electron apps would have to offer an option for this themselves.

The MiddleClickAutoscroll option worked for a while, but that no longer seems to be the case.

I think it would be possible to implement an option for this in vesktop, but I haven't done any further research.

CosmicHeron:
Middle scrolling (still) works on Plasma 6 under Wayland when launching Vesktop with these flags `--enable-blink-features=MiddleClickAutoscroll --enable-features=UseOzonePlatform --ozone-platform=wayland` without unintended pasting whatever text was in the clipboard. I have the middle click pasting disabled in the Plasma settings but iirc that settings only has an effect for native Wayland applications. I'm guessing Vesktop launches as an x11/xwayland application by default which is why just having the `--enable-blink-features=MiddleClickAutoscroll` flag is not enough.

nixRidge:
It does indeed launch as an X11 application by default. --enable-features=UseOzonePlatform isn't necessary anymore, just --ozone-platform=wayland, though it's not like leaving the first flag there hurts.

Vendicated:
I don't think this is something Vesktop should have a toggle for, as it also requires the user to disable middle click paste in their system

I documented it in [the vesktop wiki](https://vesktop.dev/linux/tips/#enable-middle-click-auto-scroll), that should be good enough

Issue #242, opened 2023-11-16
Launch as minimized


Plarpoon:
Hello, is there a way to launch the program as minimized? Even without an option in the settings but just with a launch parameter.

PR #248, opened 2023-11-18
feat: Add start as minimized toggle to settings

ZirixCZ:
This PR makes it possible for the user to start Vesktop in minimized mode. Meaning no splash screen nor the main window appearing on launch. Resolves https://github.com/Vencord/Vesktop/issues/242

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -457,6 +457,7 @@ export async function createWindows() {
 
         if (Settings.store.maximized && !isDeckGameMode) {
             mainWin!.maximize();
+            if (startMinimized) mainWin!.hide();
         }
 
         if (isDeckGameMode) {
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -453,7 +453,7 @@ export async function createWindows() {
 
     mainWin.webContents.on("did-finish-load", () => {
         splash.destroy();
-        !startMinimized && mainWin!.show();
+        if (!startMinimized) mainWin!.show();
 
         if (Settings.store.maximized && !isDeckGameMode) {
             mainWin!.maximize();
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -442,7 +442,8 @@ function createMainWindow() {
 const runVencordMain = once(() => require(join(VENCORD_FILES_DIR, "vencordDesktopMain.js")));
 
 export async function createWindows() {
-    const splash = createSplashWindow();
+    const { startMinimized } = Settings.store;
+    const splash = createSplashWindow(startMinimized);
     // SteamOS letterboxes and scales it terribly, so just full screen it
     if (isDeckGameMode) splash.setFullScreen(true);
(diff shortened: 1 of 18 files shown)


src/main/mainWindow.ts
```diff
@@ -452,7 +453,7 @@ export async function createWindows() {
 
     mainWin.webContents.on("did-finish-load", () => {
         splash.destroy();
-        mainWin!.show();
+        !startMinimized && mainWin!.show();
```
TheKodeToad, line 456:
```suggestion
        if (!startMinimized) mainWin!.show();
```

src/main/splash.ts
```diff
@@ -11,10 +11,11 @@ import { ICON_PATH, VIEW_DIR } from "shared/paths";
 
 import { Settings } from "./settings";
 
-export function createSplashWindow() {
+export function createSplashWindow(startMinimized: boolean = false) {
```
Vendicated, line 14:
```suggestion
export function createSplashWindow(startMinimized = false) {
```

Plarpoon:
Definitely interested in this one, thank you everyone!

lewisakura:
this doesn't work, the splash doesn't show but the main app does not start minimised

lewisakura:
ah, it seems like this only doesnt work if the app was maximised when it was quit. when windowing the app before quitting, it works as intended

src/main/mainWindow.ts
```diff
@@ -452,7 +453,7 @@ export async function createWindows() {
 
     mainWin.webContents.on("did-finish-load", () => {
         splash.destroy();
-        mainWin!.show();
+        if (!startMinimized) mainWin!.show();
 
         if (Settings.store.maximized && !isDeckGameMode) {
```
Vendicated, line 458: this will likely fix it
```suggestion
        if (!startMinimized && Settings.store.maximized && !isDeckGameMode) {
```

Vendicated: or better yet (thanks lewi)
```js
mainWin!.maximize();
mainWin!.hide();
```

ZirixCZ: @lewisakura  Please verify if it works now. My window manager might not have the maximize option, but during testing, when I manually set it to maximized state, it functioned correctly.

lewisakura: The underlying issue is now resolved, however this will cause a flicker which is an unintended side effect. A better solution would be to not show the window as usual, and then when the window is first brought up check if it should be maximised or not. You can use the [BrowserWindow show event](https://www.electronjs.org/docs/latest/api/browser-window#event-show) for this.

lewisakura (approved):

lewisakura:
thank you!

Vendicated:
i don't think this feature makes much sense in its current state. who would want to explicitly open vesktop and have it open minimised? when you open it with an explicit action, you want it to be usable immediately

the only case where users realistically would want vesktop to start minimised is if the action was implicit, so via the start with system option. i think this option should be reworked to be exactly that

PR #257, opened 2023-11-27
Disables always on top on updater window

zyrouge:
Always on top makes it impossible to minimize it, rendering the desktop space unusable.


Vendicated:
just close it? i don't understand why it being always on top is an issue

zyrouge:
> just close it? i don't understand why it being always on top is an issue

Because it wastes screen real estate? I don't think a update window should be blocking other applications by always being on top.

lewisakura:
then close it

PR #261, opened 2023-11-30
Notification badge in system tray icon

alchzh:
On Linux desktop, notification badges don't work for most people because there isn't a good standard. The library that Electron calls (`libunity`) is old and not many people have it. We don't have much control over the panel image either because that's controlled by the .desktop file. Personally, I find myself relying on notification badges frequently. System tray icon is easily modifiable, unlike the main panel.

This PR:
- Adds support for unread message and notification badges in the system tray
Image showing Vesktop and system tray unread message badge</img>&nbsp;&nbsp;&nbsp;&nbsp;Image showing Vesktop and system tray notification badge with number 1</img>
- Adds a corresponding setting and menu entry in Vesktop Settings
<div align="center">Image showing Vesktop settings with tray notification badge setting</img></div>

- Adds a build script to generate system tray icons with badges using badge files in `static/badges`. This allows users who wish to modify the default icon to get the badges as well.
- We need to access both `mainWin` and `tray` in `appBadge.ts` now. Moves `mainWin` and `tray` in `mainWindow.ts` to a new object `globals` to avoid circular dependencies and reordered some imports. Can revert these changes if the `require()` hacks are preferred.

I am only able to test this on Linux running KDE at the moment. Would appreciate others test this on other platforms!

Can implement the badge image creation live instead of in build step in case other modifications to the tray icon are needed (voice connected, etc.)


daliacoss:
just tested this on i3 window manager, seems to work well!

package.json
```diff
@@ -48,7 +48,10 @@
         "eslint-plugin-prettier": "^5.0.1",
         "eslint-plugin-simple-import-sort": "^10.0.0",
         "eslint-plugin-unused-imports": "^3.0.0",
+        "fast-glob": "3.3",
```
Vendicated, line 51: what's the point of adding this dependency? your use case is just `glob("./dir/*")` which could be replaced with a simple `fs.readdir`

.gitignore
```diff
@@ -1,6 +1,8 @@
-dist
+# Allow static/dist to contain empty folders
```
Vendicated, line 1: this comment is really only relevant for understanding this change, but not for future readers
```suggestion
```

scripts/build/composeTrayIcons.mts
```diff
@@ -0,0 +1,221 @@
...
+        info: validDim(info)
+    };
+}
+
+async function loadFromImageOrIco(
+    path: string | Buffer | sharp.Sharp,
+    sizeOptions?: ImageDim & { resizeICO?: boolean }
+): Promise<sharp.Sharp> {
+    if (typeof path === "string" && path.endsWith(".ico")) {
+        const icos = (await import("sharp-ico")).sharpsFromIco(path, undefined, true) as unknown as ImageData[];
```
Vendicated, line 118: what's the point of importing inline? should likely be moved to the top level

src/main/ipc.ts
```diff
@@ -15,10 +15,11 @@ import { join } from "path";
 import { debounce } from "shared/utils/debounce";
 
 import { IpcEvents } from "../shared/IpcEvents";
-import { setBadgeCount } from "./appBadge";
 import { autoStart } from "./autoStart";
 import { VENCORD_FILES_DIR, VENCORD_QUICKCSS_FILE, VENCORD_THEMES_DIR } from "./constants";
-import { mainWin } from "./mainWindow";
+import { globals } from "./mainWindow";
+// !!IMPORTANT!! ./appBadge import must occur after ./mainWindow
+import { setBadgeCount } from "./appBadge";
```
Vendicated, line 22: consider inline requiring this module to fix the circular dependency issues in a cleaner way

src/main/mainWindow.ts
```diff
@@ -34,22 +35,25 @@ import {
...
 applyDeckKeyboardFix();
 
 app.on("before-quit", () => {
     isQuitting = true;
 });
 
-export let mainWin: BrowserWindow;
+// Export a container object of objects that are used by other modules
+// but won't be initialized at import time.
+export const globals = {
```
Vendicated, line 53: why move these exports into a `global` variable? i don't really like that. i think it was better before

src/renderer/appBadge.ts
```diff
@@ -24,7 +22,8 @@ export function setBadge() {
         let totalCount = mentionCount + pendingRequests;
         if (!totalCount && hasUnread && !disableUnreadBadge) totalCount = -1;
 
-        VesktopNative.app.setBadgeCount(totalCount);
+        if (Settings.store.appBadge || Settings.store.trayBadge)
+            VesktopNative.app.setBadgeCount(totalCount, Settings.store.appBadge, Settings.store.trayBadge);
```
Vendicated, line 26: what's the point of passing these settings via ipc? the main process also has access to these settings so you don't need to do that

package.json
alchzh: Fair point, I'll refactor things

scripts/build/build.mts
```diff
@@ -51,6 +52,12 @@ async function copyVenmic() {
 
 await Promise.all([
     copyVenmic(),
+    composeTrayIcons({
```
Vendicated, line 55: as far as i understood, tray badge is not available on mac, right? if so, this copy should not happen on mac

alchzh: does building on mac imply targeting mac?

lewisakura: usually - electron-builder will use the host OS

Vendicated:
this is pretty cool, thanks for working on this!

> Can implement the badge image creation live instead of in build step in case other modifications to the tray icon are needed (voice connected, etc.)

that does sound like the most future proof / maintainable solution, i agree. just passing the arrayBuffer to the renderer, creating the image with canvas api and passing it back to node. i think that's actually also what the Discord app does

src/main/mainWindow.ts
alchzh: `tray` needs to be held in an object of some sort because it can change change or be assigned to null. I figured if I was creating a singleton for `tray` anyway might as well move `mainWindow` in there to clean up the need for require hacks elsewhere

scripts/build/build.mts
Vendicated: yes! artifacts are built on their respective platform. cross-compiling is technically supported, but
- i don't think you can cross compile mac builds
- we don't cross-compile (and don't support it)

src/renderer/appBadge.ts
alchzh: I didn't know that- will change

Vendicated:
why is mac unsupported anyway? can you not customise the menu bar icons?

src/renderer/appBadge.ts
Vendicated: you can import the settings api from src/main/settings.ts

src/main/mainWindow.ts
Vendicated: you can change let exports
```js
export let tray: BrowserWindow;

// later
tray = newWindow;
```

alchzh: Doesn't this require reimporting the module every time you want to access the up to date object? I guess I'm not actually sure how module import semantics work...

alchzh:
> why is mac unsupported anyway? can you not customise the menu bar icons?

yes but they can only contain black and alpha

src/main/mainWindow.ts
Vendicated: no, imports update dynamically. the next time you access the import it will be updated

alchzh: wow, I did not know that's what `export let` did... what the fuck. I'll change it back to the require hack as well then

src/main/ipc.ts
alchzh: I'm surprised to hear casted `require` being called cleaner than just normal imports in a typescript context... but sure

alchzh: @Vendicated to confirm since we'd need the require hacks for settings as well, you'd prefer the first version over the second version below?

JackDotJS:
any updates on this PR? would love to see this merged cus i miss notifications a lot, and there's seemingly absolutely no other indicator for new messages atm

DRAGONTOS:
+1

alchzh:
The PR was behind on some refactors in the newest versions. I've merged the changes following the refactors so it should work again now

src/main/ipc.ts
alchzh: If the current inclusion of `// eslint-disable-next-line simple-import-sort/imports` in `ipc.ts` is the roadblock on getting this merged, I'm happy to revert it to the `require` hacks. At this point I just want to not have to maintain my own branch on my system to keep up with updates...

jakjakob:
> why is mac unsupported anyway? can you not customise the menu bar icons?

macOS has it's own notification badge built in (the same in iOS), and is already used; so I think in any case it shouldn't be necessary. (I do not know the answer on your question though)

Vendicated:
tray badge was added via #1179

missing libunity will be resolved via #1180 (you still need a DE that supports the unity launcher, like KDE or "Dash to dock" on gnome)

so this pr is no longer needed. but thank you anyway!

PR #262, opened 2023-12-01
Tray: hide window if tray clicked while window focused

alchzh:
Small quality-of-life thing. Allows the user to peek at Discord via the tray without moving the mouse from the tray to the window decoration or Alt-F4ing.

Can move to a setting if this is undesirable for some users.

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -121,7 +121,13 @@ function initTray(win: BrowserWindow) {
     tray = new Tray(ICON_PATH);
     tray.setToolTip("Vesktop");
     tray.setContextMenu(trayMenu);
-    tray.on("click", () => win.show());
+    tray.on("click", () => {
+        if (win.isFocused()) {
+            win.hide();
+        } else {
+            win.show();
+        }
+    });
 
     win.on("show", () => {
         trayMenu.items[0].enabled = false;
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -126,6 +126,7 @@ function initTray(win: BrowserWindow) {
             win.hide();
         } else {
             win.show();
+            win.focus();
         }
     });
 


Vendicated:
i'm not sure this is good UX tbh. no other app does this, and i personally might spam click the tray to show vesktop

lewisakura:
this is also not a discord behaviour, so it would be inconsistent with the main app and probably confusing

lewisakura:
> I understand that it would be annoying for this to be the default if you like to make redundant clicks on the bar. Would this be acceptable as an settings option? With https://github.com/Vencord/Vesktop/pull/261 patched in I basically only use the system tray to launch the Vesktop

Maybe? But it's still a really weird option to have.

> I don't think it'd bad UX- for almost every system tray icon clicking it once opens something or does an action and clicking it again closes it or reverses the action. It's somewhat against the standard convention to have an icon do nothing once you click it once.

All of my tray icons do one action when you click them, and doesn't undo the action. The only case I've ever heard of this happening is when the tray icon *completely changes* to respect the new action (e.g., recording software). It seems like it's against convention to give a tray icon multiple actions depending on context.

> Another thought: if the window is shown but not focused / behind a bunch of other windows, clicking the tray icon should focus it (helpful for responding to dms/pings from the notification badge).

I agree, this would be useful, and in fact would be the correct behaviour for clicking the icon multiple times.

alchzh:
> All of my tray icons do one action when you click them, and doesn't undo the action. The only case I've ever heard of this happening is when the tray icon _completely changes_ to respect the new action (e.g., recording software). It seems like it's against convention to give a tray icon multiple actions depending on context.

What kind of tray icons do you have? All of my icons or widgets do something when you click them again...
image
- Notification bell: Shows notifications on first click, hides notifications on second click 
- Fcitx Input: Switches to Chinese keyboard on first click, switches back to US keyboard on second click
- Night Light: Disables nightlight temporarily on first click, re-enables night light on second click
- Volume: Opens volume slider on first click, closes volume slider on second click
- Battery: Opens battery and brightness menu on first click, closes battery and brightness menu on second click
- Bluetooth: Opens Bluetooth menu on first click, closes Bluetooth menu on second click
- etc. etc.

same applies to the buttons in the application launcher (maximize on first click, minimize on second click), the peek at desktop button (hides windows on first click, shows them again on second click), etc.

I can't find _any_ icons or buttons on the panel where the first click does something, and the second click does nothing, which is the current Vesktop behavior. Regardless of what the main Discord app does, I think the "show on click, hide on second click" behavior is absolutely standard for anything on the panel (except workspace/desktop switcher, I guess).

alchzh:
Video example of what I mean -- Vesktop is the odd one out because it doesn't do anything when I click it at the end. You can see that OBS, another app which can hide its main window into the system tray, does both show and hide when clicking on the tray icon.

https://github.com/Vencord/Vesktop/assets/8976828/0d01c758-9ab7-4bf7-8446-b1757cbb2f20

Vendicated:
please make this configurable (disabled by default)

otherwise this is not mergeable because it breaks windows convention and is inconsistent with stock discord

PR #266, opened 2023-12-03
Remove `transparent: true` for transparencyOption

Marocco2:
See https://github.com/electron/electron/issues/29937#issuecomment-1193030133

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -396,8 +396,7 @@ function createMainWindow() {
         ...(transparencyOption &&
             transparencyOption !== "none" && {
                 backgroundColor: "#00000000",
-                backgroundMaterial: transparencyOption,
-                transparent: true
+                backgroundMaterial: transparencyOption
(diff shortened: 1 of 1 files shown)


Marocco2:
Waiting for https://github.com/electron/electron/pull/40749 to be merged

Vendicated:
why are we waiting for that issue? according to the first link it should already work, no?

Marocco2:
> why are we waiting for that pr? according to the first link it should already work, no?

Because it doesn't work and glitches out (background will become black and there is heavy ghosting on UI elements)

Vendicated:
thanks! please fix ci

Vendicated:
you know you can just run autofix, riiight...

Marocco2:
> you know you can just run autofix, riiight...

Sorry about that. I repeated `pnpm test` on my end but it was throwing thousands of CR errors because I'm on Windows 😓

Thanks for autofixing for me

PR #282, opened 2023-12-09
fix: auto minimize on auto start + dbus

lewisakura:


src/main/autoStart.ts
```diff
@@ -10,38 +10,51 @@ import { join } from "path";
...
+            if (isFlatpak) {
+            } else {
+                const desktopFile = `
 [Desktop Entry]
 Type=Application
 Version=1.0
 Name=Vencord
 Comment=Vencord autostart script
-Exec=${process.execPath}
+Exec=${process.execPath} --autostart
```
Vendicated, line 37: maybe this would be slightly more self explanatory
```suggestion
Exec=${process.execPath} --is-autostart=true
```

src/main/autoStart.ts
```diff
@@ -10,38 +10,51 @@ import { join } from "path";
...
     const dir = join(configDir, "autostart");
     const file = join(dir, "vencord.desktop");
 
     return {
-        isEnabled: () => existsSync(file),
+        isEnabled: () => existsSync(file), // TODO: flatpak
+        wasAutoStarted: () => process.argv.includes("--autostart"),
         enable() {
-            const desktopFile = `
+            if (isFlatpak) {
```
Vendicated, line 29: not sure if this logic is good. why not unconditionally first try using the portal and if it doesn't work, use the desktop fallback instead? lots of DEs implement the portal so it should work on those even without flatpak, no?

lewisakura: i havent tested gdbus yet fully so this is just placeholder code

Vendicated:
could u rebase on https://github.com/Vencord/Vesktop/pull/322/files ?

Vendicated:
#1180

PR #283, opened 2023-12-09
SteamOS: add flatpak support using steam.pipe

AAGaming00:
needs `--filesystem=~/.steam` on the flatpak side

- Refactors SteamOS support to use steam.pipe instead
- Adds support for opening URLs in the steam browser in game mode
- Opens the Vesktop app page in steam instead of attempting to restart Vesktop in game mode (which won't work)

### DIFF
--- a/src/main/utils/makeLinksOpenExternally.ts
+++ b/src/main/utils/makeLinksOpenExternally.ts
@@ -7,6 +7,7 @@
 import { BrowserWindow, shell } from "electron";
 
 import { Settings } from "../settings";
+import { execSteamURL, isDeckGameMode, openURL } from "./steamOS";
 
 export function makeLinksOpenExternally(win: BrowserWindow) {
     win.webContents.setWindowOpenHandler(({ url }) => {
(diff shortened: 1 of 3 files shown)


src/main/utils/makeLinksOpenExternally.ts
```diff
@@ -7,6 +7,7 @@
 import { BrowserWindow, shell } from "electron";
 
 import { Settings } from "../settings";
+import { execSteamURL, isDeckGameMode, openURL } from "./steamOS";
```
Vendicated, line 10: it'd be good to give this func a more descriptive name
```suggestion
import { execSteamURL, isDeckGameMode, steamOpenURL } from "./steamOS";
```

AAGaming00: good point yop

src/main/utils/steamOS.ts
```diff
@@ -42,16 +39,28 @@ function getAppId(): string | null {
     return null;
 }
 
-async function execSteamURL(url: string): Promise<void> {
-    await exec(`steam -ifrunning ${url}`);
+export async function execSteamURL(url: string): Promise<void> {
+    // This doesn't allow arbitrary execution despite the weird syntax.
+    await writeFile(
+        join(process.env.HOME || "/home/deck", ".steam", "steam.pipe"),
```
Vendicated, line 45: async file writes with no synchronisation mechanisms are always a bit worrying. could multiple simultaneous writes lead to a race condition? in this case i could lead to a sandbox escape

AAGaming00: it's a pipe so I highly doubt

I've also spammed useless unnecessary shit into this pipe and steam just ignores it unless it contains that starting pattern

AAGaming00: it also always seems to execute immediately so I'm not sure if it's really async? idk

PR #304, opened 2023-12-24
[fix] Update electron to 28.1

Vam-Jam:
Updates electron, which now uses chrome 120
  - ^119 is now required for colonist due to document.hasStorageAccess()

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "@typescript-eslint/parser": "^6.13.1",
         "@vencord/types": "^0.1.2",
         "dotenv": "^16.3.1",
-        "electron": "^27.1.2",
+        "electron": "^28.1.0",
         "electron-builder": "^24.9.1",
         "esbuild": "^0.19.8",
         "eslint": "^8.54.0",


Vendicated:
how is this a fix?

what's colonist?

Vam-Jam:
My bad for not explaining. Colonist was recently added as a discord activity. (https://colonist.io/)

On the current version of Vesktop, launching Colonist will result in `Error 1003, Third party cookies must be enabled to play this activity`. 

Colonist is trying to call hasStorageAccess() (https://developer.mozilla.org/en-US/docs/Web/API/Document/hasStorageAccess), which was added in chromium v119.

Updating electron allows the activity to launch as intended.

image

Vendicated:
im confused, how does this activity work on discord desktop? it uses a much older electron / chromium

Vam-Jam:
I want to say its something to do with the user agent, it looks like the app is checking for 'discord' in the user agent before it calls hasStorageAccess. I will check whats going on shortly.

Vam-Jam:
> I want to say its something to do with the user agent, it looks like the app is checking for 'discord' in the user agent before it calls hasStorageAccess. I will check whats going on shortly.

does not seem to be the case (its just checking to display different errors based on user agent), not entirely sure how its working on the vanilla client. it does not look like hasStorageAccess is a valid func, so its not been backported.

maybe the vanilla client is being served a different file somehow?

lewisakura:
ideally we don't want to upgrade the electron version if it works on the vanilla client's older version - can we please investigate this further rather than doing a major upgrade, since it has the potential to break other things? for context, canary seems to currently be on Electron 22.3.26 so if it works fine on there then we need to work out *why*.

i mean worst comes to worst, write a polyfill of `document.hasStorageAccess = () => true` and it'll fix the game from what i'm hearing, which seems to be preferable than upgrading electron

Vendicated:
i'm closing this because you have not correctly updated the dependency (lock file also needs an update)

however, i will experiment with & test electron 28 myself and if all works correctly bump myself

thanks for bringing this issue & its fix to our attention!

Vendicated:
https://github.com/Vencord/Vesktop/commit/d71d0c1cc4914b99e0ca4fb17b32f363122bc054

PR #309, opened 2023-12-28
SteamOS: always use dark theme

AAGaming00:
The Flatpak always detects light theme, but SteamOS only has a dark mode so we can just always use dark mode there.

To the 1 person using a light theme with decky: you are insane go explode.


Vendicated:
superseded by https://github.com/Vencord/Vesktop/commit/b059516707a97be7fac2c05a67e054d0c5e73daa

PR #315, opened 2024-01-02
feat: add setting to disable hardware acceleration

zt64:


zt64:
@Vendicated rerview

Vendicated (changes requested):
LCTM

PR #316, opened 2024-01-02
feat: Add start minimized as a launch argument

redeven:
Reimplements #248 to allow starting Vencord minimized/hidden, but with an argument instead of a Settings toggle, so that it may be appended to autostart shortcuts without impacting explicit launching of the app.

Flag is `--start-minimized` to match the official Discord application's flag.

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -458,7 +458,7 @@ export async function createWindows() {
     mainWin.webContents.on("did-finish-load", () => {
         splash.destroy();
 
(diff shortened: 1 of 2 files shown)


Vendicated:
looks good. could you please also add a settings switch for `Auto start minimised`? I'm not sure how to word it, just take inspiration from Discord Windows settings

Vendicated:
actually don't worry about that, ill add it as part of https://github.com/Vencord/Vesktop/pull/322

Vendicated:
thanks

PR #322, opened 2024-01-06
autoStart: add linux portal implementation

Vendicated:
Todo:
- figure out a way to get enabled state for portal method. i tried to DBus.Properties.Get, on the Background portal but the only property is version. nothing more. we likely just need to keep a boolean in our settings or smth
- Implement --start-minimized option
- Test on various desktop environments and inside flatpak. it works for me on gnome


Vendicated:
#1180

PR #349, opened 2024-01-18
add transparency option on linux

miri999:
Added transparency option on non Windows OSes, and changed the `customTitleBar` check to only be used on windows as according to the electron docs (I misthought in the commit) transparency works on windows with borders on non Windows OSes.


Vendicated:
this is a very poor way of doing this. this setting is purely intended for windows. plain transparency should be a different unrelated setting

you can also easily check for windows inside the browser, no need for an ipc event. src/renderer/constants should already export what you need

Vendicated:
superseded by https://github.com/Vencord/Vesktop/commit/b636b65e55235fde9092843a7d3f8dae02d9d720

PR #353, opened 2024-01-20
Stop venmic on stream stop

Curve:
This PR stops venmic when the stream is ended instead of channel disconnect/switch

### DIFF
--- a/src/renderer/components/ScreenSharePicker.tsx
+++ b/src/renderer/components/ScreenSharePicker.tsx
@@ -83,11 +83,14 @@ addPatch({
 
 if (isLinux) {
     onceReady.then(() => {
-        FluxDispatcher.subscribe("VOICE_STATE_UPDATES", e => {
-            for (const state of e.voiceStates) {
-                if (state.userId === UserStore.getCurrentUser().id && state.oldChannelId && !state.channelId)
-                    VesktopNative.virtmic.stop();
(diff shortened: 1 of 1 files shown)


src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -83,11 +83,14 @@ addPatch({
 
 if (isLinux) {
     onceReady.then(() => {
-        FluxDispatcher.subscribe("VOICE_STATE_UPDATES", e => {
-            for (const state of e.voiceStates) {
-                if (state.userId === UserStore.getCurrentUser().id && state.oldChannelId && !state.channelId)
-                    VesktopNative.virtmic.stop();
+        FluxDispatcher.subscribe("STREAM_CLOSE", ({ streamKey }: { streamKey: string }) => {
```
Vendicated, line 86: does this only handle streams or could this potentially also be the camera? iirc you can both share camera and screen at once

Curve: > does this only handle streams or could this potentially also be the camera? iirc you can both share camera and screen at once

Good catch! Will investigate

Curve: Can confirm that this is only sent by Streams, not Cameras

Issue #352, opened 2024-01-20
Allow custom startup animation


jxhug:
**Is your feature request related to a problem? Please describe.**
Basically it's a little weird opening vesktop in public places lol

**Describe the solution you'd like**
A way to customize the animation by uploading a gif alternative

**Describe alternatives you've considered**
Just use the regular discord startup animation

Vendicated:
the world needs to see shiggy

zt64:
destroy shiggy

Growebis:
> destroy shiggy

I 100% agree with this.

jokes aside, i would love for it to be customizable so i can put whatever i want.

Sqaaakoi:
can we have shoggy on startup?

DissolveDZ:
totally agree, it's really easy to replace the gif in the static folder and recompile though

OboTheHobo:
> totally agree, it's really easy to replace the gif in the static folder and recompile though

for people running from source sure, most users are probably using flatpak tho. And adding an option in the app would make things easier for the end user anyway.

Vendicated:
closing in favour of https://github.com/Vencord/Vesktop/issues/667

PR #354, opened 2024-01-22
Add an option to disable the splash animation

ading2210:
This PR adds a settings option to prevent the `shiggy.gif` image from being shown on the splash screen. This setting is off by default, so it shouldn't affect any existing behavior.

image

When this setting is enabled, the splash screen instead looks like this:
image

This PR would partially solve #352.

### DIFF
--- a/src/main/splash.ts
+++ b/src/main/splash.ts
@@ -20,7 +20,7 @@ export function createSplashWindow(startMinimized = false) {
 
     splash.loadFile(join(VIEW_DIR, "splash.html"));
 
-    const { splashBackground, splashColor, splashTheming } = Settings.store;
+    const { splashBackground, splashColor, splashTheming, disableSplashAnimation } = Settings.store;
 
     if (splashTheming) {
         if (splashColor) {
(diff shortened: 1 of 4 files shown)


ading2210:
@Sqaaakoi

Would something like this be better for the window size?

image

Sqaaakoi:
> @Sqaaakoi
> 
> Would something like this be better for the window size?
> 
> image

That looks better

Vendicated:
this is just ugly and bad

ading2210:
@Vendicated How is giving users the option to disable a forced (and weird looking) animation a bad thing? 

If the only problem is that it's ugly, could you give some suggestions to make it look better?

Sqaaakoi:
> @Vendicated How is giving users the option to disable a forced (and weird looking) animation a bad thing?
> 
> If the only problem is that it's ugly, could you give some suggestions to make it look better?

disable the loading screen entirely

Sqaaakoi:
also shiggy is cute

Vendicated:
there's nothing forced about it. you're welcome to not use this piece of software

in any case, the way to make it look better is to provide alternative loading animations, or provide a way to customise it. do not use the discord icon or any other trademark/copyrighted material

ading2210:
@Vendicated Alright, I'll add a way to customize the image, and then I'll make a new PR.

PR #355, opened 2024-01-22
Add option for custom startup animations

ading2210:
Based on feedback from #354, I've implemented a way to set a custom image on the splash window. The file is loaded from the user's filesystem, so no copyrighted assets are used.

image

https://github.com/Vencord/Vesktop/assets/71154407/53f73e02-f47d-41f9-82fa-8998a8c03252

This fixes issue #352 completely.


ading2210:
I wouldn't recommend merging this at the moment, since it's currently broken under Windows.

ading2210:
Is anyone else able to test this on Windows? The person I had asked to test it before didn't rebuild the app properly. They rebuilt it today with the latest changes, and it's working now.

builtbyvys:
@ading2210 seems to work for me (i had to get something from google)
image

jacksongoode:
Works for me 👍 definitely needed!

Arian04:
is there anything blocking the merging of this that I could help with? If you guys need another tester for this on Windows in addition to the two people above, I could try it out.

src/renderer/components/Settings.tsx
```diff
@@ -153,6 +153,43 @@ export default function SettingsUi() {
                 </>
             )}
 
+            <Forms.FormTitle>Custom Spash Animation</Forms.FormTitle>
```
TheAyes, line 156:
```suggestion
            <Forms.FormTitle>Custom Splash Animation</Forms.FormTitle>
```

ading2210: This got fixed in the latest commit.
image

ading2210:
@Vendicated

Now that all the merge conflicts are fixed (thanks to @sz6084), do you have any suggestions on how this PR can be improved? Considering the demand for this feature, hopefully we can get this merged as soon as possible.

src/main/index.ts
```diff
@@ -74,6 +74,10 @@ function init() {
 
         registerScreenShareHandler();
         registerMediaPermissionsHandler();
+        //register file handler so we can load the custom splash animation from the user's filesystem
+        protocol.handle("splash-animation", () => {
+            return net.fetch("file:///"+splashAnimationPath);
```
Cynosphere, line 79: From the other PR, it would still be better if this was just a `fs.readSync` call

Cynosphere (changes requested):

src/main/index.ts
ading2210: I disagree, since the Electron protocol handler API [expects a promise for a fetch response](https://www.electronjs.org/docs/latest/api/protocol#protocolhandlescheme-handler) as the return value. 

Using a synchronous API for this isn't ideal because it will block the rest of the main process which will increase the startup time as it waits for the file to load.

The Electron docs also say to use `net.fetch` for reading local files: https://www.electronjs.org/docs/latest/breaking-changes#deprecated-protocolregisterinterceptbufferstringstreamfilehttpprotocol

Cynosphere: I was confusing it with interceptBufferProtocol the entire time then.

rklxmok:
Where are we at so I can use vencord without a furry?

Meko199:
When is this going to be added?

ilobilo:
👀

tiagorangel2011:
i know this comment is going to be marked as spam but PLEASE ADD THIS. Is there anything blocking this from being merged? it seems to be fully working. (also this serves as kind of a bump)

jxhug:
ppl want more features in the pr for some reason :/
it seems great to me

atriup:
Any idea of when this will be added? More customization is always a bonus and this has a good amount of demand for it (relative to the size of the project at least).

henry9836:
found this, thought I'd share while we wait for the developer to accept that not everyone wants to see his waifu [fork of vesktop that has the feature and is on the aur as well as other package managers] (https://github.com/verticalsync/Sunroof)

tomma365:
using ading2210's repo works for the intended feature, (I can set custom splash screen) however my only issue with this repo is the go live screen doesnt let me select an audio source from a single application like the master branch does, I just get "Entire System" or "none", my understanding is that the custom splash feature is going to be merged with master branch, but i just wanted to see if anyone had a solution to this problem

ading2210:
I added a preview of the splash image to the Vesktop settings menu. This is an important requirement for loading the animation from the Vesktop cache, since displaying only the file path won't be possible with that approach. 

vesktop_pr_1

I do eventually plan to implement all of the suggestions in #667, such as adding the ability to choose multiple images that are selected randomly and adding a toggle to turn off the splash screen entirely.

FAArik:
when will this be added ?

ading2210:
@Covkie That error is fixed in the latest commit. It was happening because Vesktop was unconditionally trying to delete and remake the splash directory, even if it didn't exist before. Obviously this would have caused an error when it tried to remove a nonexistent directory, but it didn't happen in my previous testing because that directory was already there.

lothar-cell:
Can't wait to get rid of the furry splash:)

stankovictab:
If you guys don't want to wait another half a year for this to be merged, check out [Equibop](https://github.com/Equicord/Equibop), it already has this funciton.

FAArik (approved):

ading2210:
@MMachado05 An  tag is used to load the animation, so APNG, AVIF, and WebP images should all work already.

Fxzzi:
I've gone ahead and done some testing with this PR on linux. From what I can see, everything seems to work perfectly on my end. Imo, this is ready for merging but I'm not a maintainer soooo

lewisakura:
Has this been Flatpak tested?

lothar-cell:
Why do I get the feeling that the devs cannot be bothered with this, because they love that mascot or what?

jmshgs:
freaky devs

lewisakura:
> Why do I get the feeling that the devs cannot be bothered with this, because they love that mascot or what?

I think the fact I'm asking about the test status of this PR implies I care about it, but to be frank this isn't the most important code change in the world. We are all busy people nonetheless. And of course we do love Shiggy. Who doesn't? :P

Anyhow, if anyone decides to comment on this PR going forward, please keep it to development work or testing results only since you are just sending e-mails for no reason. If people don't comply I'll lock the PR and discussion will continue on Discord instead.

ADecametre:
Amazing, thank you. :pray:

eeelbrens:
so?

kxtzownsu:
I think this is a pretty good PR, not sure why this isn't merged honestly 🤷‍♂️

SomeRandomCpu:
Please merge this. I can't open vesktop in public places until this gets merged.

LethalManBoob:
> Please merge this. I can't open vesktop in public places until this gets merged.

just use equibop

lewisakura:
> Has this been Flatpak tested?

@ading2210 I still need an update on this. This isn't mergeable if it doesn't work in Flatpak.

chrrybmb64:
> Please merge this. I can't open vesktop in public places until this gets merged.

it's a dancing pixelated anime girl, you're acting like it's nsfw

LethalManBoob:
dev is malding rn and hiding comments he doesnt like

LethalManBoob:
im glad equibop is an alternative, i refuse to use vesktop for this reason. juvenile dev

Vendicated:
Superseded by #1179 which implements this in a generic & extendable (and safer) way for both splash and tray images

Thanks for your work though!

PR #360, opened 2024-01-24
Workaround screenshare audio using microphone on debian bug

Curve:
See Vencord/venmic#15.

This is a draft because [this big boy](https://github.com/Vencord/venmic/actions/runs/7646536541/job/20835545882) is still running

### DIFF
--- a/src/main/venmic.ts
+++ b/src/main/venmic.ts
@@ -4,7 +4,7 @@
  * Copyright (c) 2023 Vendicated and Vencord contributors
  */
 
-import { PatchBay } from "@vencord/venmic";
+import type { PatchBay } from "@vencord/venmic";
 import { app, ipcMain } from "electron";
 import { join } from "path";
 import { IpcEvents } from "shared/IpcEvents";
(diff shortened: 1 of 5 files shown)


src/main/venmic.ts
```diff
@@ -51,17 +54,35 @@ ipcMain.handle(IpcEvents.VIRT_MIC_LIST, () => {
...
+    const data: LinkData = {
         include: targets.map(target => ({ key: "application.name", value: target })),
         exclude: [{ key: "application.process.id", value: getRendererAudioServicePid() }]
-    })
-);
+    };
+
+    if (workaround) {
+        data.workaround = [
+            { key: "application.process.id", value: getRendererAudioServicePid() },
```
Vendicated, line 65: nit: it would be better to store the result of this in a variable to avoid the extra work (i assume collecting metrics is kinda costly)

src/main/venmic.ts
```diff
@@ -4,11 +4,14 @@
  * Copyright (c) 2023 Vendicated and Vencord contributors
  */
 
+import { PatchBay } from "@vencord/venmic";
```
Vendicated, line 7:
```suggestion
import type { PatchBay } from "@vencord/venmic";
```

src/main/venmic.ts
Curve: True, good idea

Curve: Can we use a global cached variable?

Curve:
Tests are still failing because I had the types wrong in venmic 3.3.1, 3.3.2 is currently underway :face_with_spiral_eyes:

Vendicated:
soon

is this ready for merge?

Vendicated (approved):

Curve:
> soon
> 
> is this ready for merge?

Yes

PR #370, opened 2024-01-31
Add zoom slider, match official client zoom

Kaydax:
This code is really bad, so I expect many change request. However its working fine.

This adds a zoom slider to the Vesktop settings page as well as rebinds ctrl+/- to use these zoom controls instead of the built in electron zoom controls.


src/main/mainWindow.ts
```diff
@@ -470,6 +503,32 @@ export async function createWindows() {
                 mainWin!.maximize();
             }
         });
+
+        mainWin.on("focus", () => {
```
Vendicated, line 507: this is a really bad way to do this. why not use menu items with accelerators like previously?

Vendicated:
i dont think this is very useful. why would anyone do this instead of just using zoom shortcuts?

sunnniee:
> i dont think this is very useful. why would anyone use this instead of just using zoom shortcuts?

because they simply don't know the zoom shortcuts exist. changing zoom is a reoccurring question in the support channel

Vendicated:
people also constantly ask how to change vencord location in vesktop even tho its inside vesktop settings

people are simply stupid. you cant fix that by adding pointless settings

Kaydax:
> i dont think this is very useful. why would anyone use this instead of just using zoom shortcuts?

I explained this in the dev channel. The current zoom shortcut buttons do not match the same zoom scale as the official discord client, thus causing totally different zoom levels between the two. It's more of a OCD thing than anything, but also makes the client feel more like the official one

src/main/mainWindow.ts
Kaydax: Because it refused to work with the menu items after hours of trying. If anyone is able to get it working that way instead that would be super helpful

DRAGONTOS:
+1

TomTheDragon:
The zoom shortcut does not seem to work on linux (Vesktop and/or official client)
So a slider like the official client has, seems to be the only way to get zoom functionality

D3SOX:
@TomTheDragon I don't why it doesn't work for you but <kbd>CTRL</kbd> + <kbd>0</kbd> / <kbd>-</kbd> / <kbd>+</kbd> work fine for me on Linux with KDE Plasma

TomTheDragon:
> @TomTheDragon I don't why it doesn't work for you but CTRL + 0 / - / + work fine for me on Linux with KDE Plasma
I just gave it a try on my other machine and zoom over the shortcut works fine over there.
Both machines run KDE6, so not sure whats the problem then.
But hey it at least shows that you can run into the problem, in which you are dependent on a slider.

TomTheDragon:
> @TomTheDragon I don't why it doesn't work for you but CTRL + 0 / - / + work fine for me on Linux with KDE Plasma

I just gave it a try on my other machine and zoom over the shortcut works fine over there.
Both machines run KDE6, so not sure whats the problem then.
But hey it at least shows that you can run into the problem, in which you are dependent on a slider.

Kaydax:
So any updates on this? If there is still a problem with how I handle the inputs for zoom, it would be good to help find a better way if that's the only issue.

Geryythekid:
> So any updates on this? If there is still a problem with how I handle the inputs for zoom, it would be good to help find a better way if that's the only issue.

I have a problem, zoom simply does not work, i am on kde6. Besides that i only have a 60% keyboard and there way to make those combos work. I dont wanna ask a buddy for a full sized keyboard every time i need to change smth like this. Getting a full sized keyboard is not a solution either.

Kaydax:
> I have a problem, zoom simply does not work, i am on kde6. Besides that i only have a 60% keyboard and there is no way to make those combos work. Getting a full sized keyboard is not a solution either, nor having 2 keyboards on the desk at the same time.

Wdym the "no way to make those combos work". The only key combos used are CTRL +/-. and 60% keyboards have +/-

Geryythekid:
> > I have a problem, zoom simply does not work, i am on kde6. Besides that i only have a 60% keyboard and there is no way to make those combos work. Getting a full sized keyboard is not a solution either, nor having 2 keyboards on the desk at the same time.
> 
> Wdym the "no way to make those combos work". The only key combos used are CTRL +/-. and 60% keyboards have +/-

Those combos work only in english layout on the 60%keyboard, which i rarely use. Took me some time to find out.

src/main/mainWindow.ts
Vendicated: it works very much the same. did you maybe forget to remove the default zoom handlers? they might take priority over your accelerators

src/renderer/components/settings/WindowZoom.tsx
```diff
@@ -0,0 +1,48 @@
...
+    return (
+        <>
+            <Forms.FormTitle className={Margins.top16 + " " + Margins.bottom8}>Zoom Level</Forms.FormTitle>
+            <Slider
+                className={Margins.top20}
+                initialValue={zoomFactor}
+                defaultValue={1}
+                onValueChange={v => {
+                    settings.zoomFactor = v;
+                    VesktopNative.win.zoom(v);
```
Vendicated, line 37: you don't need this ipc call. you can register a settings change listener in the native process, look at the `initSettingsListener` function in mainWindow.ts

src/renderer/components/settings/WindowZoom.tsx
```diff
@@ -0,0 +1,48 @@
...
+                    setZoomFactor(v);
```
Vendicated, line 38: you don't need this state. via `useSettings()`, you can get a stateful settings object that will automatically rerender your component on change

the parent component already uses this hook so you should not need to call it yourself

src/renderer/components/settings/WindowZoom.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Forms, Slider, useEffect, useState } from "@vencord/types/webpack/common";
+
+import { SettingsComponent } from "./Settings";
+
+export const WindowZoom: SettingsComponent = ({ settings }) => {
+    const [zoomFactor, setZoomFactor] = useState(settings.zoomFactor ?? 1);
+
+    useEffect(() => {
+        const handleZoomChange = event => {
+            console.log("zoom changed", event.detail);
```
Vendicated, line 17:
```suggestion
```

src/renderer/components/settings/WindowZoom.tsx
```diff
@@ -0,0 +1,48 @@
...
+            setZoomFactor(event.detail);
+        };
+
+        window.addEventListener("zoomChanged", handleZoomChange);
```
Vendicated, line 21: you shouldn't need this event. changing settings will already update your state & rerender the ui

Vendicated (changes requested):
you convinced me so i guess we can go forward and get this merged! 

however there are some things that need to be fixed first

src/main/mainWindow.ts
Vendicated: it works very much the same. did you maybe forget to remove the default zoom handlers? they might take priority over your accelerators

src/shared/settings.d.ts
```diff
@@ -20,6 +20,7 @@ export interface Settings {
     arRPC?: boolean;
     appBadge?: boolean;
     disableMinSize?: boolean;
+    zoomFactor?: number;
```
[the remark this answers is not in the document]
Kaydax: Then how would it persist after restarting?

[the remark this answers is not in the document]
Vendicated: state is also persisted!

src/renderer/components/settings/WindowZoom.tsx
Kaydax: I removed the listener, and the UI element does not update when the zoom gets changed, even with a listener

NoltoxGit:
Hi here!
Any news about this PR? I rly need this option bcs shortcuts doesn't work too.

PR #326, opened 2024-01-07
feat: Add global shortcuts

tuxinal:
~~i'm using electron's globalShortcuts api here which doesn't have a way to detect tell the difference between keyup and keydown. this means the push to talk/mute/deafen/etc actions don't really work. we'd probably have to switch to something else at some point~~
currently implementing [venbind](https://github.com/tuxinal/venbind) to use instead of electron's api 

resolves #18

PR #371, opened 2024-02-01
Global shortcuts support

D3SOX:
This is pretty bare-bones for now and doesn't work as Electron IPC events don't allow a function as parameter, but I'm already opening this here just for reference


D3SOX:
I just noticed I wasted my time. Closing in favor of #326

PR #404, opened 2024-02-23
Tray: Add left click hide/show feature

0bCdian:
This pr adds the option to click on the tray to hide or show the window.

I made sure it:

- [x]  adds the corresponding entry to the vesktop settings tab.
- [x]  is optional and disabled by default. 

I'm making this pull request since #262 seems to be stale.


DRAGONTOS:
+1

src/main/mainWindow.ts
```diff
@@ -73,6 +73,13 @@ const [addSettingsListener, removeSettingsListeners] = makeSettingsListenerHelpe
 const [addVencordSettingsListener, removeVencordSettingsListeners] = makeSettingsListenerHelpers(VencordSettings);
 
 function initTray(win: BrowserWindow) {
+    const trayClickHandler = Settings.store.clickTrayToShowHide
```
Vendicated, line 76: if you move this ternary inside the callback, it wont require a restart:
```js
const onTrayClick = () => {
	if (Settings.store.clickTrayToShowHide && win.isVisible())
		win.hide();
	else
		win.show();
}
```

0bCdian: Hi! Thanks for reviewing my pr, I have already committed your proposed changes (I don't know how I didn't think of your way of doing it first, it's a much cleaner solution to such a simple feature), sorry for the delay, I don't know how the notification slipped without me noticing.

Vendicated:
thank you!

PR #431, opened 2024-03-10
Call BrowserWindow.isMaximized and subscribe to related events from renderer

Sqaaakoi:
I'm building a plugin that effectively replaces Discord's custom titlebar, and I am using this to make the titlebar more compact when the window is maximized. Other plugins could find this useful, or even Vesktop itself could patch the standard maximized button to have a different icon.


Vendicated:
i'd just do this via a native.ts file instead. don't really see a use case for this other than your plugin

PR #432, opened 2024-03-10
Fix runtime error on startup

ZirixCZ:
Fixes runtime error on startup when network's unavailable

#### Before

https://github.com/Vencord/Vesktop/assets/49836430/7c17cae7-647b-4958-86e1-d7eacd039ec5

#### After

https://github.com/Vencord/Vesktop/assets/49836430/25da02f8-ddc7-41cd-8a40-60dda520fc84

### DIFF
--- a/src/main/utils/http.ts
+++ b/src/main/utils/http.ts
@@ -23,7 +23,7 @@ export async function downloadFile(url: string, file: string, options: RequestOp
 export function simpleReq(
     url: string,
     options: RequestOptions = {},
-    retries = 0,
+    retries = 10,
     backoffDelay = 1000
 ): Promise<IncomingMessage> {
     return new Promise<IncomingMessage>((resolve, reject) => {
@@ -33,10 +33,10 @@ export function simpleReq(
             if (statusCode! >= 300) return simpleReq(headers.location!, options).then(resolve).catch(reject);
             resolve(res);
         }).on("error", err => {
-            if (retries > 10) {
+            if (retries < 0) {
                 reject(err);
             } else {
-                setTimeout(() => resolve(simpleReq(url, options, retries + 1, backoffDelay * 2)), backoffDelay);
+                setTimeout(() => resolve(simpleReq(url, options, retries - 1, backoffDelay * 2)), backoffDelay);
             }
         });
     });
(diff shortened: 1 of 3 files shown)


src/main/utils/http.ts
```diff
@@ -20,26 +20,40 @@ export async function downloadFile(url: string, file: string, options: RequestOp
...
             if (statusCode! >= 400) return void reject(`${statusCode}: ${statusMessage} - ${url}`);
             if (statusCode! >= 300) return simpleReq(headers.location!, options).then(resolve).catch(reject);
-
             resolve(res);
+        }).on("error", err => {
+            if (retries > 10) {
+                reject(err);
+            } else {
+                setTimeout(() => resolve(simpleReq(url, options, retries + 1, backoffDelay * 2)), backoffDelay);
+            }
```
rushiiMachine, line 40: Shouldn't this be instead recursing backwards until 0?
The `retries` parameter would instead default to 10.

src/main/utils/http.ts
```diff
@@ -20,26 +20,40 @@ export async function downloadFile(url: string, file: string, options: RequestOp
...
-        res.once("error", reject);
-        res.on("data", chunk => chunks.push(chunk));
-        res.once("end", () => resolve(Buffer.concat(chunks)));
-    });
+            res.once("error", reject);
+            res.on("data", chunk => chunks.push(chunk));
+            res.once("end", () => resolve(Buffer.concat(chunks)));
+        });
+    } catch (e) {
+        console.log(e);
```
rushiiMachine, line 57:
```suggestion
        console.error(e);
```

rushiiMachine (changes requested):

src/main/utils/http.ts
ZirixCZ: Refactored it https://github.com/Vencord/Vesktop/pull/432/commits/eaabc3aae9a8fc44f1887c5dc4c4a2537e190515

Vendicated:
thanks but like i said, we should also switch to the fetch api

thus superseded by https://github.com/Vencord/Vesktop/commit/97267ef89a24f4fefae256481bf9bd8380a1d3d4

PR #439, opened 2024-03-12
Vennie splash

TheKodeToad:
Epic PR

credit to shoritsu (probably should wait for permission ig?)


ZirixCZ:
This could be fixed by https://github.com/Vencord/Vesktop/pull/355 with keeping shiggy as the default

Sqaaakoi (changes requested):
i wouldn't replace shiggy entirely, instead i would make it randomly pick shiggy or vennie

TheKodeToad:
Shiggy was just used without permission so I don't think it should stay in Vesktop and Vencord - we have no rights to use it (ig an emoji is a bit different though)

lewisakura:
Was permission granted?

Vendicated:
closing in favour of custom assets

PR #445, opened 2024-03-15
Allow for higher stream FPS

Grafaffel:
Allow for higher stream FPS in the Picker

### DIFF
--- a/src/renderer/components/ScreenSharePicker.tsx
+++ b/src/renderer/components/ScreenSharePicker.tsx
@@ -24,7 +24,7 @@ import { addPatch } from "renderer/patches/shared";
 import { isLinux, isWindows } from "renderer/utils";
 
 const StreamResolutions = ["480", "720", "1080", "1440"] as const;
-const StreamFps = ["15", "30", "60"] as const;
+const StreamFps = ["15", "30", "60", "120"] as const;
 
 const MediaEngineStore = findStoreLazy("MediaEngineStore");
 


PolisanTheEasyNick:
Discord will not allow streaming more than 60 FPS

kaitlynkittyy:
I second that, ive tried and to no avail, but even then the bitrate barely holds 1080p60 stable, 1440 is really rough on it, so even 480p120 would be really tough.

Grafaffel:
how does `FakeNitor` does it then?

Grafaffel:
I saw a custom plugin in the Vencord server by someone, wich would allow for higher fps streaming. He showed it in a VC, and is was silky smoooth

PolisanTheEasyNick:
I applied 120 fps at Track settings, and we still are hardcapped to 60 fps, even when sometimes we reach for moment >60 fps
image
Also, we have strange bitrate behavior.

Grafaffel:
I found the plugin now: [Better Screenshare](https://github.com/philhk/Vencord/tree/better-plugins/src/plugins/betterScreenshare.desktop)

lewisakura:
There is no purpose if it doesn't actually increase the framerate, as indicated by some people in this post

Grafaffel:
> There is no purpose if it doesn't actually increase the framerate, as indicated by some people in this post

ok, I'm closing the issue now

Issue #191, opened 2023-10-27
Screenshare has poor performance due to no hardware acceleration


MiMillieuh:
Hello, 

That's a native issue too but I may have a suggestion to make it work.

When you're streaming on discord (at least on wayland via xwayland or pipewire for now) the framerate of the app or game are extremely slow. But when discord is the focused window it's smooth again.

My suggestion would be to have a toggle (via a plugin or directly integrated) to keep the screenshare preview without having the window focused. (Disabled by default because I don't know it this issue affect windows and mac users. but it's not the case on the native client for windows at least.)

But that's only a suggestion and maybe that's not even what causes the issue... but it's weird that when the window is focused, this issue isn't happening.

MiMillieuh:
> I've noticed only vesktop struggles with terrible screenshare performance (minutes per frame slow). Native discord only experiences random frame drops.

Well honestly I have the exact same issue on native client... 

Using OBS camera works fine tho

maotovisk:
I'm also not able to stream at a decent frame rate, 60fps does not work and it does not even seem to be remotely close to 30.

MrShelbySann:
Same here, I can't even stream in 30, it's like a slideshow. The issue is not present on the native client.

Vendicated:
no. its on by default

MiMillieuh:
I've tried to mess with hardware acceleration but it's still a slideshow...

s-talay:
I think the native client for linux sucks because of electron not being able to utilize hardware acceleration? I don't know Electron all too well but maybe, since this project uses Gluon, there might be other ways of solving this issue?

Vendicated:
we don't use gluon.

ant0n-0x0000:
it can use hardware acceleration also

s-talay:
> we don't use gluon.

I dont know how I possibly miss-read the readme this bad lol

MiMillieuh:
Honestly I would like to try to fake having the discord window always on focus. 
I've noticed that when I click on my discord window and that the stream shows on the client (not showing the black thing to "use less ressources") it's way more smooth.

But is it possible to prevent that screen and function to come ?

vickyorlo:
[...]

Issue #480, opened 2024-04-02
Add support for screensharing in gamescope using the gamescope pipewire video source.


ILOVEPIE:
**Describe the solution you'd like.**
If we can consume the gamescope:out_0 pipewire video source we can use that for screensharing under gamescope.

kaitlynkittyy:
I'd like to add enhancements where possible, but I'm lacking an understanding of this issue. 
Could you please explain in detail what use case this would have and why it would be important?

ILOVEPIE:
> And also please specify how we can reproduce it.
> Because today I tried starting game with gamescope and it just started in another window, which I really can choose using portal picker or I can just stream whole display with game window.

Gamescope is a wayland compositor that runs by itself when it is not being launched from another compositor. It does not support a portal capture at the moment and only has the pipewire output. The problem is that this is the main compositor for the Steam Decks game mode, which means the only way to capture video in the Steam Decks game mode of a game is to use the pipewire output.

kaitlynkittyy:
> > And also please specify how we can reproduce it.
> > Because today I tried starting game with gamescope and it just started in another window, which I really can choose using portal picker or I can just stream whole display with game window.
> 
> Gamescope is a wayland compositor that runs by itself when it is not being launched from another compositor. It does not support a portal capture at the moment and only has the pipewire output. The problem is that this is the main compositor for the Steam Deck's game mode, which means the only way to capture video in the Steam Deck's game mode of a game is to use the pipewire output.

ah, i see
that will most likely not be supported in this PR, but will be investigated in the future.

PolisanTheEasyNick:
I believe for achieving this we can use [org.freedesktop.portal.ScreenCast.OpenPipeWireRemote](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html#org-freedesktop-portal-screencast-openpipewireremote) DBus call.
But for achieving streams creation directly from DBus requires whole streaming code refactor and migrating from electron's `desktopCapturer.getSources` calling to direct DBus calls for Linux systems, which is a real major of work.

simvux:
As far as I know you can have multiple implementations of the same portal exposed via dbus. So, maybe this could be fixed by having an xdg-desktop-portal implementation for gamescope, thus this being an issue for upstream gamescope?

ILOVEPIE:
[...]

Issue #463, opened 2024-03-20
Vesktop streams resolution of source window no matter what


BuyMyMojo:
**Describe the bug**

When screen sharing Vesktop will ignore your prefered resolution and use the resolution of the display/app you are using. I confirmed this with a friend and they are receiving a full 4K stream even tho discord reports that I chose 720p for both of us.

**To Reproduce**

1. Start stream
2. Click on `Voice Connected` in the bottom left corner
3. Click `Debug`
4. Under the title `RTC DEBUG: STREAM` there will be a tab named `Outbound` go there
5. Check the `Resolution` reported there
![image](https://github.com/Vencord/Vesktop/assets/46948241/ee574799-23bc-4bb5-a416-bfd3db0a67de)

**Expected behavior**

scale video to 720p during encode

**Screenshots**

![Screenshot_20240320_224946](https://github.com/Vencord/Vesktop/assets/46948241/21f503c6-437d-4c1f-b7a9-4e64de9eec22)

I am streaming the file explorer window behind the debug window. it is in one corner of my 4K monitor and I chose to stream in discord at 480p:
![image](https://github.com/Vencord/Vesktop/assets/46948241/a7af71f5-5139-4e6c-9da9-1a1e3493a42c)
![image](https://github.com/Vencord/Vesktop/assets/46948241/bd511511-c081-44c8-ad18-fc35ebd57a6a)

**Desktop (please complete the following information):**
 - OS/Distro: NixOS
 - Desktop Environment (linux only): KDE Plasma 5
 - Version: 23.11 (But I'm running nix on the unstable channel)

**Command line output**

nothing to really note

```
vesktop 
[arRPC > ipc] listening at /run/user/1000/discord-ipc-0
[arRPC > websocket] listening on 6463
[arRPC > process] started
Error occurred in handler for 'VencordGetUpdates': Error: No handler registered for 'VencordGetUpdates'
    at WebContents.<anonymous> (node:electron/js2c/browser_init:2:78280)
    at WebContents.emit (node:events:514:28)

```
![image](https://github.com/Vencord/Vesktop/assets/46948241/d90bfa52-27e9-45d4-930a-f77712e04149)

**Additional context**

PolisanTheEasyNick:
I found that for some reason, framerate takes directly from mediaDevices.getDisplayMedia (which returns [MediaStream](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream)) object.
[...]

PR #489, opened 2024-04-07
overhaul & improve Linux screenshare

kaitlynkittyy:
basically what the title says. streaming is now smooth on linux. finally.

TODO before merge: 
- [x] Fix resolution and framerate selection
- [x] Fix bug that stops users from changing the streamed window (created by allowing resolution changes)
- [x] Fix bug where changing windows allows you to lower the fps but not raise it
- [x] Beautify modal for starting a stream(show preview and content type as a dropdown) 
- [x] Fix issue where stream quality overlay does not update to show the current quality of the running stream
- [x] Enable ``contentHint`` and allow the user to choose between "motion" and "detail"
- [x] Allow users to start streams while watching another stream

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
         "arrpc": "github:OpenAsar/arrpc#98879cae0565e6fce34e4cb6f544bf42c6a7e7c8"
     },
     "optionalDependencies": {
-        "@vencord/venmic": "^3.3.2"
+        "@vencord/venmic": "^3.3.3"
     },
     "devDependencies": {
         "@fal-works/esbuild-plugin-global-externals": "^2.1.2",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
         "arrpc": "github:OpenAsar/arrpc#98879cae0565e6fce34e4cb6f544bf42c6a7e7c8"
     },
     "optionalDependencies": {
-        "@vencord/venmic": "^3.3.3"
+        "@vencord/venmic": "^3.4.1"
     },
     "devDependencies": {
         "@fal-works/esbuild-plugin-global-externals": "^2.1.2",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "@typescript-eslint/parser": "^7.2.0",
         "@vencord/types": "^0.1.2",
         "dotenv": "^16.4.5",
-        "electron": "29.1.6",
+        "electron": "^29.1.1",
         "electron-builder": "^24.13.3",
         "esbuild": "^0.20.1",
         "eslint": "^8.57.0",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "@typescript-eslint/parser": "^7.2.0",
         "@vencord/types": "^0.1.2",
         "dotenv": "^16.4.5",
-        "electron": "^29.1.1",
+        "electron": "29.1.6",
         "electron-builder": "^24.13.3",
         "esbuild": "^0.20.1",
         "eslint": "^8.57.0",
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -27,10 +27,12 @@ process.env.VENCORD_USER_DATA_DIR = DATA_DIR;
 function init() {
     const { disableSmoothScroll, hardwareAcceleration } = Settings.store;
 
-    if (hardwareAcceleration === false) app.disableHardwareAcceleration();
-    if (hardwareAcceleration === true) {
+    if (hardwareAcceleration === false) {
+        app.disableHardwareAcceleration();
+    } else {
         app.commandLine.appendSwitch("enable-features", "VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,VaapiVideoDecoder");
     }
+
     if (disableSmoothScroll) {
         app.commandLine.appendSwitch("disable-smooth-scrolling");
     }
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -36,18 +36,11 @@ function init() {
     app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
     app.commandLine.appendSwitch(
         "enable-features",
-        "UseOzonePlatform,VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,WebRTCPipeWireCapturer,VaapiVideoDecoder,CanvasOopRasterization,Vulkan"
+        "UseOzonePlatform,VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,WebRTCPipeWireCapturer,VaapiVideoDecoder"
     );
     app.commandLine.appendSwitch("use-vulkan");
     app.commandLine.appendSwitch("use-gl", "angle");
     app.commandLine.appendSwitch("use-angle", "gl");
-    app.commandLine.appendSwitch("enable-gpu-rasterization");
-    app.commandLine.appendSwitch("enable-gpu");
-    app.commandLine.appendSwitch("enable-unsafe-webgpu");
-    app.commandLine.appendSwitch("enable-oop-rasterization");
-    app.commandLine.appendSwitch("enable-native-gpu-memory-buffers");
-    app.commandLine.appendSwitch("enable-accelerated-video");
-    app.commandLine.appendSwitch("enable-zero-copy");
     app.commandLine.appendSwitch("ignore-gpu-blocklist");
     app.commandLine.appendSwitch("ozone-platform-hint", "wayland");
     // WinRetrieveSuggestionsOnlyOnDemand: Work around electron 13 bug w/ async spellchecking on Windows.
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -36,7 +36,7 @@ function init() {
     app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
     app.commandLine.appendSwitch(
         "enable-features",
-        "UseOzonePlatform,VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,VaapiVP8Encoder,VaapiVP9Encoder,VaapiAV1Encoder,WebRTCPipeWireCapturer,VaapiVideoDecoder,CanvasOopRasterization,Vulkan"
+        "UseOzonePlatform,VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,WebRTCPipeWireCapturer,VaapiVideoDecoder,CanvasOopRasterization,Vulkan"
     );
     app.commandLine.appendSwitch("use-vulkan");
     app.commandLine.appendSwitch("use-gl", "angle");
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -34,22 +34,15 @@ function init() {
 
     // work around chrome 66 disabling autoplay by default
     app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
-    app.commandLine.appendSwitch(
-        "enable-features",
-        "UseOzonePlatform,VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,WebRTCPipeWireCapturer,VaapiVideoDecoder"
-    );
-    app.commandLine.appendSwitch("use-vulkan");
-    app.commandLine.appendSwitch("use-gl", "angle");
-    app.commandLine.appendSwitch("use-angle", "gl");
-    app.commandLine.appendSwitch("ignore-gpu-blocklist");
-    app.commandLine.appendSwitch("ozone-platform-hint", "wayland");
+    app.commandLine.appendSwitch("enable-features", "VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,VaapiVideoDecoder");
+    app.commandLine.appendSwitch("ozone-platform-hint", "auto");
     // WinRetrieveSuggestionsOnlyOnDemand: Work around electron 13 bug w/ async spellchecking on Windows.
     // HardwareMediaKeyHandling,MediaSessionService: Prevent Discord from registering as a media service.
     //
     // WidgetLayering (Vencord Added): Fix DevTools context menus https://github.com/electron/electron/issues/38790
     app.commandLine.appendSwitch(
         "disable-features",
-        "WinRetrieveSuggestionsOnlyOnDemand,HardwareMediaKeyHandling,MediaSessionService,WidgetLayering,UseChromeOSDirectVideoDecoder"
+        "WinRetrieveSuggestionsOnlyOnDemand,HardwareMediaKeyHandling,MediaSessionService,WidgetLayering"
     );
 
     // In the Flatpak on SteamOS the theme is detected as light, but SteamOS only has a dark mode, so we just override it
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -34,7 +34,30 @@ function init() {
 
     // work around chrome 66 disabling autoplay by default
     app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
-
+    app.commandLine.appendSwitch(
+        "enable-features",
+        "UseOzonePlatform,VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,VaapiVP8Encoder,VaapiVP9Encoder,VaapiAV1Encoder,WebRTCPipeWireCapturer,VaapiVideoDecoder,CanvasOopRasterization,Vulkan"
+    );
+    app.commandLine.appendSwitch("use-vulkan");
+    app.commandLine.appendSwitch("use-gl", "angle");
+    app.commandLine.appendSwitch("use-angle", "gl");
+    app.commandLine.appendSwitch("enable-gpu-rasterization");
+    app.commandLine.appendSwitch("enable-gpu");
+    app.commandLine.appendSwitch("enable-unsafe-webgpu");
+    app.commandLine.appendSwitch("enable-oop-rasterization");
+    app.commandLine.appendSwitch("enable-native-gpu-memory-buffers");
+    app.commandLine.appendSwitch("enable-accelerated-video");
+    app.commandLine.appendSwitch("enable-zero-copy");
+    app.commandLine.appendSwitch("ignore-gpu-blocklist");
+    app.commandLine.appendSwitch("ozone-platform-hint", "wayland");
+    // WinRetrieveSuggestionsOnlyOnDemand: Work around electron 13 bug w/ async spellchecking on Windows.
+    // HardwareMediaKeyHandling,MediaSessionService: Prevent Discord from registering as a media service.
+    //
+    // WidgetLayering (Vencord Added): Fix DevTools context menus https://github.com/electron/electron/issues/38790
+    app.commandLine.appendSwitch(
+        "disable-features",
+        "WinRetrieveSuggestionsOnlyOnDemand,HardwareMediaKeyHandling,MediaSessionService,WidgetLayering,UseChromeOSDirectVideoDecoder"
+    );
     // WinRetrieveSuggestionsOnlyOnDemand: Work around electron 13 bug w/ async spellchecking on Windows.
     // HardwareMediaKeyHandling,MediaSessionService: Prevent Discord from registering as a media service.
     //
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -28,13 +28,15 @@ function init() {
     const { disableSmoothScroll, hardwareAcceleration } = Settings.store;
 
     if (hardwareAcceleration === false) app.disableHardwareAcceleration();
+    if (hardwareAcceleration === true) {
+        app.commandLine.appendSwitch("enable-features", "VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,VaapiVideoDecoder");
+    }
     if (disableSmoothScroll) {
         app.commandLine.appendSwitch("disable-smooth-scrolling");
     }
 
     // work around chrome 66 disabling autoplay by default
     app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
-    app.commandLine.appendSwitch("enable-features", "VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,VaapiVideoDecoder");
     // WinRetrieveSuggestionsOnlyOnDemand: Work around electron 13 bug w/ async spellchecking on Windows.
     // HardwareMediaKeyHandling,MediaSessionService: Prevent Discord from registering as a media service.
     //
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -35,7 +35,6 @@ function init() {
     // work around chrome 66 disabling autoplay by default
     app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
     app.commandLine.appendSwitch("enable-features", "VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,VaapiVideoDecoder");
-    app.commandLine.appendSwitch("ozone-platform-hint", "auto");
     // WinRetrieveSuggestionsOnlyOnDemand: Work around electron 13 bug w/ async spellchecking on Windows.
     // HardwareMediaKeyHandling,MediaSessionService: Prevent Discord from registering as a media service.
     //
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -51,14 +51,6 @@ function init() {
         "disable-features",
         "WinRetrieveSuggestionsOnlyOnDemand,HardwareMediaKeyHandling,MediaSessionService,WidgetLayering,UseChromeOSDirectVideoDecoder"
     );
-    // WinRetrieveSuggestionsOnlyOnDemand: Work around electron 13 bug w/ async spellchecking on Windows.
-    // HardwareMediaKeyHandling,MediaSessionService: Prevent Discord from registering as a media service.
-    //
-    // WidgetLayering (Vencord Added): Fix DevTools context menus https://github.com/electron/electron/issues/38790
-    app.commandLine.appendSwitch(
-        "disable-features",
-        "WinRetrieveSuggestionsOnlyOnDemand,HardwareMediaKeyHandling,MediaSessionService,WidgetLayering"
-    );
 
     // In the Flatpak on SteamOS the theme is detected as light, but SteamOS only has a dark mode, so we just override it
     if (isDeckGameMode) nativeTheme.themeSource = "dark";
--- a/src/renderer/components/ScreenSharePicker.tsx
+++ b/src/renderer/components/ScreenSharePicker.tsx
@@ -88,6 +88,7 @@ addPatch({
             bitrateMax: 8000000,
             bitrateTarget: 600000
         });
+        console.log("opts: ", opts);
 //         Object.assign(opts.encode, {
 //             framerate,
 //             pixelCount: height * width
@@ -95,7 +96,8 @@ addPatch({
         Object.assign(opts.capture, {
             framerate,
             width,
-            height
+            height,
+            pixelCount: height * width
         });
     }
 });
@@ -447,11 +449,12 @@ function ModalComponent({
                             console.log("No current stream.");
                         }
                         try {
-                            submit({
-                                id: selected!,
-                                ...settings
-                            });
-                            if (conn) {
+                            if(!conn) {
+                                submit({
+                                    id: selected!,
+                                    ...settings
+                                });
+                            } else {
                                 console.log(conn);
                                 const track = conn.input.stream.getVideoTracks()[0];
                                 console.log(track);
--- a/src/renderer/components/ScreenSharePicker.tsx
+++ b/src/renderer/components/ScreenSharePicker.tsx
@@ -16,7 +16,6 @@ import {
     Select,
     Switch,
     Text,
-    Tooltip,
     UserStore,
     useState
 } from "@vencord/types/webpack/common";
@@ -188,31 +187,26 @@ function StreamSettings({
             deps: [source.id]
         }
     );
-    const preview = (
-        <div className="vcd-screen-picker-preview">
-            <img src={thumb} alt="stream preview" />;
-        </div>
-    );
+    const [visible, setVisible] = useState(false);
     // the source's name is not properly being displayed
     return (
         <div>
             <Forms.FormTitle>What you're streaming</Forms.FormTitle>
             <section>
-                <div className="vcd-screen-picker-tooltip">
-                    <Card className={"vcd-screen-picker-card"}>
-                        <Tooltip text={preview}>
-                            {({ onMouseEnter, onMouseLeave }) => (
-                                <div
-                                    className="vcd-screen-picker-tooltip"
-                                    onMouseEnter={onMouseEnter}
-                                    onMouseLeave={onMouseLeave}
-                                >
-                                    Show Preview
-                                </div>
-                            )}
-                        </Tooltip>
-                    </Card>
-                </div>
+                <Card className={"vcd-screen-picker-card vcd-screen-picker-preview-buttons"}>
+                    <button
+                        className="vcd-screen-picker-subtle-button"
+                        onClick={() => {
+                            setVisible(!visible);
+                        }}
+                    >
+                        Show Preview
+                    </button>
+                    <button className="vcd-screen-picker-button">Change</button>
+                </Card>
+                <Card className={visible ? "vcd-screen-picker-card vcd-screen-picker-preview fade-in" : "not-visible"}>
+                    <img src={thumb} alt="stream preview" />
+                </Card>
             </section>
             <Forms.FormTitle>Stream Settings</Forms.FormTitle>
             <Card className="vcd-screen-picker-card">
@@ -415,34 +409,58 @@ function ModalComponent({
             </Modals.ModalContent>
 
             <Modals.ModalFooter className="vcd-screen-picker-footer">
-                <div className="vcd-screen-picker-footerbuttons">
-                    <Button
-                        color={Button.Colors.TRANSPARENT}
-                        className="vcd-screen-picker-startbuttons"
-                        onClick={() => {
-                            close();
-                            navigator.mediaDevices.getDisplayMedia();
-                        }}
-                    >
-                        Go Back
-                    </Button>
-                    <Button
-                        className="vcd-screen-picker-startbuttons"
-                        disabled={!selected}
-                        onClick={() => {
-                            currentSettings = settings;
-                            // If there are 2 connections, the second one is the existing stream.
-                            // In that case, we patch its quality
-                            try {
-                                var conn = [...MediaEngineStore.getMediaEngine().connections].find(
-                                    connection => connection.streamUserId === UserStore.getCurrentUser().id
-                                );
-                            } catch {
-                                console.log("No current stream.");
+                <Button
+                    disabled={!selected}
+                    onClick={() => {
+                        currentSettings = settings;
+                        // If there are 2 connections, the second one is the existing stream.
+                        // In that case, we patch its quality
+                        try {
+                            var conn = [...MediaEngineStore.getMediaEngine().connections].find(
+                                connection => connection.streamUserId === UserStore.getCurrentUser().id
+                            );
+                        } catch {
+                            console.log("No current stream.");
+                        }
+                        console.log([...MediaEngineStore.getMediaEngine().connections]);
+                        try {
+                            if (conn) {
+                                console.log(conn);
+                                const track = conn.input.stream.getVideoTracks()[0];
+                                console.log(track);
+                                const frameRate = Number(settings.fps);
+                                const height = Number(settings.resolution);
+                                const width = Math.round(height * (16 / 9));
+                                var constraints = track.getConstraints();
+                                const newConstraints = {
+                                    ...constraints,
+                                    frameRate,
+                                    advanced: [{ width: width, height: height }],
+                                    resizeMode: "none"
+                                };
+                                track.applyConstraints(newConstraints).then(() => {
+                                    console.log("Applied constraints from ScreenSharePicker successfully.");
+                                    console.log("New constraints:", track.getConstraints());
+                                });
+
+                                // changing stream quality description
+                                conn.videoStreamParameters[0].maxFrameRate = Number(settings.fps);
+                                conn.videoStreamParameters[0].maxResolution.height = Number(settings.resolution);
+                                conn.videoStreamParameters[0].maxResolution.width = Math.round(height * (16 / 9));
                             }
-                            console.log([...MediaEngineStore.getMediaEngine().connections]);
-                            try {
-                                if (conn) {
+                        } catch {
+                            console.log("No current stream.");
+                        }
+                        try {
+                            submit({
+                                id: selected!,
+                                ...settings
+                            });
+
+                            // reapply contraints after some time to let discord resubmit stream
+                            // i believe there MUST be way to do it cleaner..
+                            if (conn) {
+                                setTimeout(() => {
                                     console.log(conn);
                                     const track = conn.input.stream.getVideoTracks()[0];
                                     console.log(track);
@@ -465,69 +483,27 @@ function ModalComponent({
                                     conn.videoStreamParameters[0].maxFrameRate = Number(settings.fps);
                                     conn.videoStreamParameters[0].maxResolution.height = Number(settings.resolution);
                                     conn.videoStreamParameters[0].maxResolution.width = Math.round(height * (16 / 9));
-                                }
-                            } catch {
-                                console.log("No current stream.");
+                                }, 100);
                             }
-                            try {
-                                submit({
-                                    id: selected!,
-                                    ...settings
-                                });
+                        } catch {
+                            console.log("Unable to start stream.");
+                        }
 
-                                // reapply contraints after some time to let discord resubmit stream
-                                // i believe there MUST be way to do it cleaner..
-                                if (conn) {
-                                    setTimeout(() => {
-                                        console.log(conn);
-                                        const track = conn.input.stream.getVideoTracks()[0];
-                                        console.log(track);
-                                        const frameRate = Number(settings.fps);
-                                        const height = Number(settings.resolution);
-                                        const width = Math.round(height * (16 / 9));
-                                        var constraints = track.getConstraints();
-                                        const newConstraints = {
-                                            ...constraints,
-                                            frameRate,
-                                            advanced: [{ width: width, height: height }],
-                                            resizeMode: "none"
-                                        };
-                                        track.applyConstraints(newConstraints).then(() => {
-                                            console.log("Applied constraints from ScreenSharePicker successfully.");
-                                            console.log("New constraints:", track.getConstraints());
-                                        });
+                        close();
+                    }}
+                >
+                    Go Live
+                </Button>
 
-                                        // changing stream quality description
-                                        conn.videoStreamParameters[0].maxFrameRate = Number(settings.fps);
-                                        conn.videoStreamParameters[0].maxResolution.height = Number(
-                                            settings.resolution
-                                        );
-                                        conn.videoStreamParameters[0].maxResolution.width = Math.round(
-                                            height * (16 / 9)
-                                        );
-                                    }, 100);
-                                }
-                            } catch {
-                                console.log("Unable to start stream.");
-                            }
-
-                            close();
-                        }}
-                    >
-                        Go Live
+                {selected && !skipPicker ? (
+                    <Button color={Button.Colors.TRANSPARENT} onClick={() => setSelected(void 0)}>
+                        Back
                     </Button>
-                </div>
-                <div className="vcd-screen-picker-cancel">
-                    {selected && !skipPicker ? (
-                        <Button color={Button.Colors.TRANSPARENT} onClick={() => setSelected(void 0)}>
-                            Back
-                        </Button>
-                    ) : (
-                        <Button color={Button.Colors.TRANSPARENT} onClick={close}>
-                            Cancel
-                        </Button>
-                    )}
-                </div>
+                ) : (
+                    <Button color={Button.Colors.TRANSPARENT} onClick={close}>
+                        Cancel
+                    </Button>
+                )}
             </Modals.ModalFooter>
         </Modals.ModalRoot>
     );
--- a/src/renderer/components/ScreenSharePicker.tsx
+++ b/src/renderer/components/ScreenSharePicker.tsx
@@ -187,31 +187,27 @@ function StreamSettings({
             deps: [source.id]
         }
     );
-    const notVisible = false;
-    function setVisible() {
-        return !notVisible;
-    }
+    const [visible, setVisible] = useState(false);
     // the source's name is not properly being displayed
     return (
         <div>
             <Forms.FormTitle>What you're streaming</Forms.FormTitle>
-            <Switch
-                onChange={setVisible}
-                value={notVisible ?? false}
-                className="vcd-screen-picker-audio vcd-screen-picker-preview-switch"
-            >
-                Show Preview
-            </Switch>
-            <Card
-                className={
-                    notVisible
-                        ? "vcd-screen-picker-card vcd-screen-picker-preview not-visible"
-                        : "vcd-screen-picker-card vcd-screen-picker-preview"
-                }
-            >
-                <img src={thumb} alt="stream preview" />
-                <Text variant="text-sm/normal">{source.name}</Text>
-            </Card>
+            <section>
+                <Card className={"vcd-screen-picker-card vcd-screen-picker-preview-buttons"}>
+                    <button
+                        className="vcd-screen-picker-subtle-button"
+                        onClick={() => {
+                            setVisible(!visible);
+                        }}
+                    >
+                        Show Preview
+                    </button>
+                    <button className="vcd-screen-picker-button">Change</button>
+                </Card>
+                <Card className={visible ? "vcd-screen-picker-card vcd-screen-picker-preview fade-in" : "not-visible"}>
+                    <img src={thumb} alt="stream preview" />
+                </Card>
+            </section>
             <Forms.FormTitle>Stream Settings</Forms.FormTitle>
             <Card className="vcd-screen-picker-card">
                 <div className="vcd-screen-picker-quality">
--- a/src/renderer/components/ScreenSharePicker.tsx
+++ b/src/renderer/components/ScreenSharePicker.tsx
@@ -53,6 +53,9 @@ interface Source {
 
 export let currentSettings: StreamSettings | null = null;
 
+const logger = new Vencord.Util.Logger("VesktopScreenShare");
+var isOverloaded = false;
+
 addPatch({
     patches: [
         {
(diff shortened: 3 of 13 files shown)


kaitlynkittyy:
FINALLY
im bad at regex

NextWork123:
Seem having issues with these things i compile with your pull request and i start and it crash i tried too to start on terminal and i have an error "SIGSEGV (Address boundary error)" with terminal it loading and when it loads two second and crash again :/ 

i only merged your pull on the fork i do

package.json
```diff
@@ -37,7 +37,7 @@
         "@typescript-eslint/parser": "^7.2.0",
         "@vencord/types": "^0.1.2",
         "dotenv": "^16.4.5",
-        "electron": "^29.1.1",
+        "electron": "29.1.6",
```
Vendicated, line 40: ?

src/main/index.ts
```diff
@@ -34,7 +34,23 @@ function init() {
...
+    app.commandLine.appendSwitch("ignore-gpu-blocklist");
+    app.commandLine.appendSwitch("ozone-platform-hint", "wayland");
+    // WinRetrieveSuggestionsOnlyOnDemand: Work around electron 13 bug w/ async spellchecking on Windows.
+    // HardwareMediaKeyHandling,MediaSessionService: Prevent Discord from registering as a media service.
+    //
+    // WidgetLayering (Vencord Added): Fix DevTools context menus https://github.com/electron/electron/issues/38790
+    app.commandLine.appendSwitch(
+        "disable-features",
+        "WinRetrieveSuggestionsOnlyOnDemand,HardwareMediaKeyHandling,MediaSessionService,WidgetLayering,UseChromeOSDirectVideoDecoder"
+    );
```
Vendicated, line 53: looks like you accidently duplicated this line

package.json
kaitlynkittyy: i didnt even notice i did that lol, was probably a remnant from testing. Only index.ts and ScreenSharePicker.tsx are necessary as far as I know

IlChitarrista:
Okay, I'm testing this, the original theory I had on https://github.com/Vencord/Vesktop/issues/191#issuecomment-2030694090 seems to be correct.

So, what this MR does is remove the artificial limit to 2500kbps.  
On my system the limit, as shown by the Bitrate(Target) in the debug menu, is not updated to 80.000kbps but only 4623 with some variability.
This improves things as now 1440p can hold 10-20fps and 1080p almost locked to 30 but it's still being throttled at some other level.

The workaround (moving the cursor) to temporarily prompt updates still gives some result but it's not needed anymore on my system, at 1080p.

The next step would be figuring out why the resolution selector is not working properly as setting the resolution to one that can be handled and supported by most clients, especially when not using nitro, would totally solve this on sufficiently powerful connections.

Vendicated:
> On my system the limit, as shown by the Bitrate(Target) in the debug menu, is not updated to 80.000kbps but only 4623 with some variability.

perhaps because it can only keep up with that much bitrate? iirc hardware encoding may not work with nvidia gpus, so you may be stuck with software encoding and thus have a bottleneck

IlChitarrista:
Also, on a secondary note, there seems to be something off about multimonitor setups on Wayland.
If I disable the secondary monitor I get ~25fps at 1440p. It seems that, on multimonitors, the primary one gets somehow throttled when there isn't cursor movement.
Lastly, I think we should also look into [contentHint](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/contentHint).

Still, once this is ready, it will be a nice step in the right direction.

kaitlynkittyy:
> > On my system the limit, as shown by the Bitrate(Target) in the debug menu, is not updated to 80.000kbps but only 4623 with some variability.
> 
> perhaps because it can only keep up with that much bitrate? iirc hardware encoding may not work with nvidia gpus, so you may be stuck with software encoding and thus have a bottleneck

I've just been able to recreate the same problem. I don't exactly know what's caused it, but it's definitely not limited to one client.

IlChitarrista:
> Also, on a secondary note, there seems to be something off about multimonitor setups on Wayland. If I disable the secondary monitor I get ~25fps at 1440p. It seems that, on multimonitors, the primary one gets somehow throttled when there isn't cursor movement. Lastly, I think we should also look into [contentHint](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/contentHint).
> 
> Still, once this is ready, it will be a nice step in the right direction.

The cursor issue isn't present when streaming an individual app.
Also, the resolution is properly set when streaming an individual app.

src/main/index.ts
```diff
@@ -34,14 +34,22 @@ function init() {
 
     // work around chrome 66 disabling autoplay by default
     app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
-
+    app.commandLine.appendSwitch(
+        "enable-features",
+        "UseOzonePlatform,VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,WebRTCPipeWireCapturer,VaapiVideoDecoder"
```
Vendicated, line 39: these should be obsolete
```suggestion
        "VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,VaapiVideoDecoder"
```

src/main/index.ts
```diff
@@ -34,14 +34,22 @@ function init() {
...
+    );
+    app.commandLine.appendSwitch("use-vulkan");
+    app.commandLine.appendSwitch("use-gl", "angle");
+    app.commandLine.appendSwitch("use-angle", "gl");
+    app.commandLine.appendSwitch("ignore-gpu-blocklist");
+    app.commandLine.appendSwitch("ozone-platform-hint", "wayland");
```
Vendicated, line 45:
```suggestion
```

src/main/index.ts
```diff
@@ -34,14 +34,22 @@ function init() {
...
     app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
-
+    app.commandLine.appendSwitch(
+        "enable-features",
+        "UseOzonePlatform,VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,WebRTCPipeWireCapturer,VaapiVideoDecoder"
+    );
+    app.commandLine.appendSwitch("use-vulkan");
+    app.commandLine.appendSwitch("use-gl", "angle");
+    app.commandLine.appendSwitch("use-angle", "gl");
+    app.commandLine.appendSwitch("ignore-gpu-blocklist");
```
Vendicated, line 44: why?

src/main/index.ts
```diff
@@ -34,14 +34,22 @@ function init() {
...
     // work around chrome 66 disabling autoplay by default
     app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
-
+    app.commandLine.appendSwitch(
+        "enable-features",
+        "UseOzonePlatform,VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,WebRTCPipeWireCapturer,VaapiVideoDecoder"
+    );
+    app.commandLine.appendSwitch("use-vulkan");
+    app.commandLine.appendSwitch("use-gl", "angle");
+    app.commandLine.appendSwitch("use-angle", "gl");
```
Vendicated, line 43: please verify that these 3 switches are actually necessary

src/main/index.ts
```diff
@@ -34,14 +34,22 @@ function init() {
...
+    app.commandLine.appendSwitch("ignore-gpu-blocklist");
+    app.commandLine.appendSwitch("ozone-platform-hint", "wayland");
     // WinRetrieveSuggestionsOnlyOnDemand: Work around electron 13 bug w/ async spellchecking on Windows.
     // HardwareMediaKeyHandling,MediaSessionService: Prevent Discord from registering as a media service.
     //
     // WidgetLayering (Vencord Added): Fix DevTools context menus https://github.com/electron/electron/issues/38790
     app.commandLine.appendSwitch(
         "disable-features",
-        "WinRetrieveSuggestionsOnlyOnDemand,HardwareMediaKeyHandling,MediaSessionService,WidgetLayering"
+        "WinRetrieveSuggestionsOnlyOnDemand,HardwareMediaKeyHandling,MediaSessionService,WidgetLayering,UseChromeOSDirectVideoDecoder"
```
Vendicated, line 52: why?

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -59,6 +59,21 @@ addPatch({
...
+        {
+            find: "x-google-max-bitrate",
+            replacement: {
+                // eslint-disable-next-line no-useless-escape
+                match: /"x-google-max-bitrate=".concat\(\i\)/,
+                replace: '"x-google-max-bitrate=".concat("80_000")'
+            }
+        },
+        {
+            find: "x-google-max-bitrate",
```
Vendicated, line 72: replacement takes either one replacement or an array of replacements

instead of having two patches with the same find, merge them into one patch

kaitlynkittyy:
I still want to try and raise the bitrate, as its somehow locked at 4696kbps right now. and during one of my tests it randomly jumped to 8000kbps

kaitlynkittyy:
> I still want to try and raise the bitrate, as its somehow locked at 4696kbps right now. and during one of my tests it randomly jumped to 8000kbps

this has since been resolved

andrepapoti:
How are you able to monitor bitrate during streaming ? I'm still facing a few issues with FPS in this branch but it's already better than the current `main` branch

Vendicated:
go into src/main/mainWindow.ts and slap this under initArRPC:
```js
const w = new BrowserWindow();
w.loadURL("chrome://webrtc-internals");
```

kaitlynkittyy:
> go into src/main/mainWindow.ts and slap this under initArRPC:
> 
> ```js
> const w = new BrowserWindow();
> w.loadURL("chrome://webrtc-internals");
> ```

that's one method, but not the one I use. If you click Video Connected above your profile name (where you would normally mute and shit in the bottom left) and then click debug.

lennoxlotl:
I wonder if discord has some kind of limit for framerates in the browser. It seems like its capped at 30 fps, no matter what you change.

kaitlynkittyy:
> I wonder if discord has some kind of limit for framerates in the browser. It seems like its capped at 30 fps, no matter what you change.

I've been able to raise the framerate above 30, but that's going to take a lot more work before its at a point where resolution and framerate selections are functioning properly.

lennoxlotl:
> > I wonder if discord has some kind of limit for framerates in the browser. It seems like its capped at 30 fps, no matter what you change.
> 
> I've been able to raise the framerate above 30, but that's going to take a lot more work before its at a point where resolution and framerate selections are functioning properly.

Ah I see, I have tried multiple things and never got it to anything above 30 but at the same time I am quite a big noob with discords api

rurigk:
I tested this, my friends just say its looking like a normal discord screenshare, not a slideshow anymore

Mato1111:
Works well for me too

andrepapoti:
Working great on my side too :)

Unrelated question: Is there any way to get AV1 encoding atm ?

kaitlynkittyy:
> Working great on my side too :)
> 
> Unrelated question: Is there any way to get AV1 encoding atm ?

Not as of right now, and it's not a priority because it would take a good bit of work, and getting streams to a functional state is my priority. If you'd like to try and implement it yourself though, go ahead :)

kaitlynkittyy:
I'm working on getting some stuff ready to merge all at once, hopefully this can be a single merge that fixes this and resolution selections. There's a bit more work to be done to fix that, and then there's also a weird issue where as soon as someone starts watching a stream, the bitrate is limited to 4696kbps.

lennoxlotl:
For me it caps at around 8000kbps when someone starts watching. Is it maybe resolution based?

kaitlynkittyy:
> For me, I do not have bitrate capping to 4696kpbs even when someone is watching a stream image
> 
> When streaming something with amount of motion (ufo test on all screen), then bitrate increases up to 8mbps~, but framerate drops to 15~ image Streamed in 1920x1080

are you on wayland? do you have nitro? what gpu do you have? i need as much information as i can get because even on a fresh clone of my branch i get locked to 4696 when someone starts watching a stream.

PolisanTheEasyNick:
> > For me, I do not have bitrate capping to 4696kpbs even when someone is watching a stream image
> > When streaming something with amount of motion (ufo test on all screen), then bitrate increases up to 8mbps~, but framerate drops to 15~ image Streamed in 1920x1080
> 
> are you on wayland? do you have nitro? what gpu do you have? i need as much information as i can get because even on a fresh clone of my branch i get locked to 4696 when someone starts watching a stream.

Arch Linux, KDE, Wayland, nvidia-dkms and I do have nitro

PolisanTheEasyNick:
> > > > For me, I do not have bitrate capping to 4696kpbs even when someone is watching a stream image
> > > > When streaming something with amount of motion (ufo test on all screen), then bitrate increases up to 8mbps~, but framerate drops to 15~ image Streamed in 1920x1080
> > > 
> > > 
> > > are you on wayland? do you have nitro? what gpu do you have? i need as much information as i can get because even on a fresh clone of my branch i get locked to 4696 when someone starts watching a stream.
> > 
> > 
> > Arch Linux, KDE, Wayland, nvidia-dkms and I do have nitro Also need to note that I edited screenShareAudio.ts as previously described here: [#463 (comment)](https://github.com/Vencord/Vesktop/issues/463#issuecomment-2041652755), but for 60 fps and full hd (even if resolution settings are ignored in some reason)
> 
> hmm, Im working on applying the same settings inside of the screensharepicker.tsx file, but i still get locked at 4696. It's interesting, I'll have to test even more.

If I can help somehow or give you more info/logs you can text me here or dm in discord @ob3r0n

PolisanTheEasyNick:
Also, i have another idea how to probably increase stability of stream. If we somehow will get RTCRtpSender object of stream then we can set [degradationPreference](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/getParameters#degradationpreference) property to "maintain-framerate", and it maybe will fix problem with fps drops on low bandwidth instead of decreasing video quality (just like default discord streams?)

Calosis:
> Also, i have another idea how to probably increase stability of stream. If we somehow will get RTCRtpSender object of stream then we can set [degradationPreference](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/getParameters#degradationpreference) property to "maintain-framerate", and it maybe will fix problem with fps drops on low bandwidth instead of decreasing video quality (just like default discord streams?)

I think this is a good idea. I tried applying this myself, but couldn't manage to do so. So cannot confirm if it has any effect

Calosis:
> Also, i have another idea how to probably increase stability of stream. If we somehow will get RTCRtpSender object of stream then we can set [degradationPreference](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/getParameters#degradationpreference) property to "maintain-framerate", and it maybe will fix problem with fps drops on low bandwidth instead of decreasing video quality (just like default discord streams?)

I think this is a good idea. I tried applying this myself, but couldn't manage to do so. So cannot confirm if it has any effect

kaitlynkittyy:
> Also, i have another idea how to probably increase stability of stream. If we somehow will get RTCRtpSender object of stream then we can set [degradationPreference](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/getParameters#degradationpreference) property to "maintain-framerate", and it maybe will fix problem with fps drops on low bandwidth instead of decreasing video quality (just like default discord streams?)

this is deprecated, I am not 100% sure but I believe you're now supposed to use contentHint, which I am somewhat trying to apply

kaitlynkittyy:
TODO before merge: 
- [ ] Fix resolution and framerate selection
- [ ] Enable ``contentHint`` and allow the user to choose between "motion" and "text"
- [ ] (maybe) enable VP9 codec for supported GPUs

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -228,7 +229,39 @@ function StreamSettings({
...
+                    <Forms.FormTitle>Content Type</Forms.FormTitle>
+                    <div>
+                        <div className="vcd-screen-picker-radios">
+                            <label className="vcd-screen-picker-radio" data-checked={settings.contentHint === "motion"}>
+                                <Text variant="text-sm/bold">Prefer Smoothness</Text>
+                                <input
+                                    type="radio"
+                                    name="contenthint"
+                                    value="motion"
+                                    checked={settings.contentHint === "moiton"}
```
D3SOX, line 242: Just a quick heads-up there's a typo in `motion` here ^^

andrepapoti:
A bit unrelated but when testing this PR I found an issue when streaming Monster Hunter: World, all other games I've tried run fine except this one, but I can't figure out why.
This may be worth taking a look because it could be affecting other applications as well.

PolisanTheEasyNick:
> A bit unrelated but when testing this PR I found an issue when streaming Monster Hunter: World, all other games I've tried run fine except this one, but I can't figure out why.
> This may be worth taking a look because it could be affecting other applications as well.
> 
> Are there any logs that I can provide that would be useful for investigating it ?

How exactly issue represents? Stream not starting? Or updates only when you move mouse in window? If last, you can try to stream game window instead of whole screen

kaitlynkittyy:
> @PolisanTheEasyNick sorry, I've added context to the original comment. The framerate is abysmally low.
> 
> I've tried running the game in windowed mode and full screen, and I also tried streaming both the game window and the whole screen, the issue persisted in all of the possible permutations involving these configurations

Could you do me a favor and open the debug menu and show what that menu looks like?

to open the debug menu, you need to:
1. click on Voice connected in the bottom left of vesktop
2. Press debug in the popup
3. choose the outbound connection for your RTC stream

andrepapoti:
@kaitlynkittyy
The issue w/ MH:W
image

kaitlynkittyy:
> @kaitlynkittyy The debug info when replicating the issue with MH:W The "input frame rate" at 1 is very suspect but I don't know what it refers to, my game is running normally image

It's odd that somehow the game updates aren't triggering the stream to increase its framerate. I need to look into figuring out how to lock the framerate as high as possible, even when there's no action

EmptyVisual:
It seems that the screenshare function is cropping the output, but it's not adjusting the display to fit a 1080p canvas. 

To illustrate, when I share a game running at 3440x1440 (21:9) resolution and set the screenshare to 1080p at 60fps, the image gets overcropped. However, in the official Discord client, the canvas is correctly scaled down to fit the selected resolution.

Screenshot example:
814D8379-124D-4E4A-A765-59D1FC659C53_1_101_o

PolisanTheEasyNick:
> It seems that the screenshare function is cropping the output, but it's not adjusting the display to fit a 1080p canvas.
> 
> To illustrate, when I share a game running at 3440x1440 (21:9) resolution and set the screenshare to 1080p at 60fps, the image gets overcropped. However, in the official Discord client, the canvas is correctly scaled down to fit the selected resolution.
> 
> Screenshot example: 814D8379-124D-4E4A-A765-59D1FC659C53_1_101_o

Did you tested it with latest commit?

kaitlynkittyy:
i dont know what changed, but something changed.
Streams are now limited to 720p30 and a few bugs that were fixed are now back.
We're working on resolving that.

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -88,10 +88,10 @@ addPatch({
             bitrateMax: 8000000,
             bitrateTarget: 600000
         });
-        Object.assign(opts.encode, {
-            framerate,
-            pixelCount: height * width
-        });
+//         Object.assign(opts.encode, {
```
Cynosphere, line 92:
```suggestion
        if (opts?.encode)
          Object.assign(opts.encode, {
            framerate,
            pixelCount: height * width
          });
```

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -424,7 +424,7 @@ function ModalComponent({
                         }
                         console.log([...MediaEngineStore.getMediaEngine().connections]);
                         try {
-                            if (conn.streamUserId === UserStore.getCurrentUser().id) {
+                            if (conn) {
```
Cynosphere, line 429:
```suggestion
                            if (conn?.streamUserId === UserStore.getCurrentUser().id) {
```

Cynosphere (changes requested):

src/renderer/components/ScreenSharePicker.tsx
PolisanTheEasyNick: Do we really need this? Because conn variable is exists only when we previously find it by comparing IDs:
```
var conn = [...MediaEngineStore.getMediaEngine().connections].find(
  connection => connection.streamUserId === UserStore.getCurrentUser().id
);
```

SpyTec: With how it's implemented currently it is needed as conn is defined within a try-catch. If it were to raise we wouldn't have `conn`

PolisanTheEasyNick: Maybe we can just write `if(conn?)`? I just want to not do unnecessary === calls

SpyTec: Ah you're right as conn will already be defined as the user's stream. Yeah I think just using `if (conn)` would be enough

src/renderer/components/ScreenSharePicker.tsx
PolisanTheEasyNick: fixed this in [006f4e9](https://github.com/Vencord/Vesktop/pull/489/commits/006f4e96dea21fee018e446e84a19a92ddabbd1e)

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -347,29 +417,70 @@ function ModalComponent({
...
+                        }
+                        try {
+                            submit({
+                                    id: selected!,
+                                    ...settings
+                                });
+
+                            //reapply contraints after some time to let discord resubmit stream
+                            //i believe there MUST be way to do it cleaner..
+                            setTimeout(() => {
```
Cynosphere, line 461: `submit` is more than likely async.
```suggestion
                            submit({
                              id: selected!,
                              ...settings
                            }).then(() => {
```

Cynosphere (changes requested):

src/renderer/components/ScreenSharePicker.tsx
PolisanTheEasyNick: I retested with .then() and it doesn't work. Seems like discord doing screenres+30fps stuff after then() call. So I think that we should use timeout :c

Curve:
> iirc hardware encoding may not work with nvidia gpus

I've used this in the past: https://github.com/elFarto/nvidia-vaapi-driver

regalialong:
> > iirc hardware encoding may not work with nvidia gpus
> 
> I've used this in the past: [elFarto/nvidia-vaapi-driver](https://github.com/elFarto/nvidia-vaapi-driver)

From Readme:
> Hardware decoding only, encoding is [not supported](https://github.com/elFarto/nvidia-vaapi-driver/issues/116).

Curve:
> > > iirc hardware encoding may not work with nvidia gpus
> > 
> > 
> > I've used this in the past: [elFarto/nvidia-vaapi-driver](https://github.com/elFarto/nvidia-vaapi-driver)
> 
> From Readme:
> 
> > Hardware decoding only, encoding is [not supported](https://github.com/elFarto/nvidia-vaapi-driver/issues/116).

Oh, I missed that, thanks :D

kaitlynkittyy:
> > > > iirc hardware encoding may not work with nvidia gpus
> > > 
> > > 
> > > I've used this in the past: [elFarto/nvidia-vaapi-driver](https://github.com/elFarto/nvidia-vaapi-driver)
> > 
> > From Readme:
> > 
> > > Hardware decoding only, encoding is [not supported](https://github.com/elFarto/nvidia-vaapi-driver/issues/116).
> 
> Oh, I missed that, thanks :D

according to the archwiki, it should be - [ ] supported by the normal nvidia driver/nvidia-utils

https://wiki.archlinux.org/title/Hardware_video_acceleration#NVIDIA_driver_only

Sqaaakoi:
> > > > > iirc hardware encoding may not work with nvidia gpus
> > > > 
> > > > 
> > > > I've used this in the past: [elFarto/nvidia-vaapi-driver](https://github.com/elFarto/nvidia-vaapi-driver)
> > > 
> > > 
> > > From Readme:
> > > > Hardware decoding only, encoding is [not supported](https://github.com/elFarto/nvidia-vaapi-driver/issues/116).
> > 
> > 
> > Oh, I missed that, thanks :D
> 
> according to the archwiki, it should be supported by the normal nvidia driver/nvidia-utils
> 
> https://wiki.archlinux.org/title/Hardware_video_acceleration#NVIDIA_driver_only
> 
> keep in mind as of right now vesktop uses h264

afaik novideo and chromium on wayland do not mix well :trollface:

kaitlynkittyy:
> > > > > > iirc hardware encoding may not work with nvidia gpus
> > > > > 
> > > > > 
> > > > > I've used this in the past: [elFarto/nvidia-vaapi-driver](https://github.com/elFarto/nvidia-vaapi-driver)
> > > > 
> > > > 
> > > > From Readme:
> > > > > Hardware decoding only, encoding is [not supported](https://github.com/elFarto/nvidia-vaapi-driver/issues/116).
> > > 
> > > 
> > > Oh, I missed that, thanks :D
> > 
> > according to the archwiki, it should be supported by the normal nvidia driver/nvidia-utils
> > 
> > https://wiki.archlinux.org/title/Hardware_video_acceleration#NVIDIA_driver_only
> > 
> > keep in mind as of right now vesktop uses h264
> 
> afaik novideo and chromium on wayland do not mix well :trollface: 



fair point lmao, hopefully explicit sync can help majorly

kaitlynkittyy:
I believe this Resolves #191 and #463

Calosis:
I'm not sure if any of you are having the same issue I am finding now, but hardware acceleration (encoding) doesn't seem to work anymore (?), aka, nvtop doesn't show encode anymore. Both myself, and a friend, are having this issue. We are both on Fedora 40, could someone else confirm this?

PolisanTheEasyNick:
>  I'm not sure if any of you are having the same issue I am finding now, but hardware acceleration (encoding) doesn't seem to work anymore (?), aka, nvtop doesn't show encode anymore. Both myself, and a friend, are having this issue. We are both on Fedora 40, could someone else confirm this?

Does Linux even supports hardware encoding? I can be wrong at this point.

lennoxlotl:
> > I'm not sure if any of you are having the same issue I am finding now, but hardware acceleration (encoding) doesn't seem to work anymore (?), aka, nvtop doesn't show encode anymore. Both myself, and a friend, are having this issue. We are both on Fedora 40, could someone else confirm this?
> 
> Does Linux even supports hardware encoding? I can be wrong at this point.

Yes it does using vaapi

kaitlynkittyy:
> > > I'm not sure if any of you are having the same issue I am finding now, but hardware acceleration (encoding) doesn't seem to work anymore (?), aka, nvtop doesn't show encode anymore. Both myself, and a friend, are having this issue. We are both on Fedora 40, could someone else confirm this?
> > 
> > Does Linux even supports hardware encoding? I can be wrong at this point.
> 
> Yes it does using vaapi (only on Intel and AMD GPUs)

According to archwiki, it's also supported on Nvidia, as NVENC supports h264, which is what vesktop uses 
whether or not it actually applies and works with chromium, I don't know

kaitlynkittyy:
> > > > > I'm not sure if any of you are having the same issue I am finding now, but hardware acceleration (encoding) doesn't seem to work anymore (?), aka, nvtop doesn't show encode anymore. Both myself, and a friend, are having this issue. We are both on Fedora 40, could someone else confirm this?
> > > > 
> > > > 
> > > > Does Linux even supports hardware encoding? I can be wrong at this point.
> > > 
> > > 
> > > Yes it does using vaapi (only on Intel and AMD GPUs)
> > 
> > According to archwiki, it's also supported on Nvidia, as NVENC supports h264, which is what vesktop uses whether or not it actually applies and works with chromium, I don't know
> 
> I don't believe it will work on Chromium, they don't implement NVENC (?), at least according to arch wiki:
> image
> 

If I'm not mistaken, you can use NVENC as a backend for either VA-API or VDPAU, If it's VDPAU, you can use that as a backend for VA-API

Calosis:
> > > > > > I'm not sure if any of you are having the same issue I am finding now, but hardware acceleration (encoding) doesn't seem to work anymore (?), aka, nvtop doesn't show encode anymore. Both myself, and a friend, are having this issue. We are both on Fedora 40, could someone else confirm this?
> > > > > 
> > > > > 
> > > > > Does Linux even supports hardware encoding? I can be wrong at this point.
> > > > 
> > > > 
> > > > Yes it does using vaapi (only on Intel and AMD GPUs)
> > > 
> > > 
> > > According to archwiki, it's also supported on Nvidia, as NVENC supports h264, which is what vesktop uses whether or not it actually applies and works with chromium, I don't know
> > 
> > 
> > I don't believe it will work on Chromium, they don't implement NVENC (?), at least according to arch wiki:
> > image
> 
> If I'm not mistaken, you can use NVENC as a backend for either VA-API or VDPAU, If it's VDPAU, you can use that as a backend for VA-API

Are you referring to this? https://github.com/elFarto/nvidia-vaapi-driver -- If so, this only supports decode. I would love to know how, if possible, encode can work for Chromium. Will help a friend with an Nvidia gpu a bit. This is off-topic for this PR anyway :))

kaitlynkittyy:
> > > > > > > I'm not sure if any of you are having the same issue I am finding now, but hardware acceleration (encoding) doesn't seem to work anymore (?), aka, nvtop doesn't show encode anymore. Both myself, and a friend, are having this issue. We are both on Fedora 40, could someone else confirm this?
> > > > > > 
> > > > > > 
> > > > > > Does Linux even supports hardware encoding? I can be wrong at this point.
> > > > > 
> > > > > 
> > > > > Yes it does using vaapi (only on Intel and AMD GPUs)
> > > > 
> > > > 
> > > > According to archwiki, it's also supported on Nvidia, as NVENC supports h264, which is what vesktop uses whether or not it actually applies and works with chromium, I don't know
> > > 
> > > 
> > > I don't believe it will work on Chromium, they don't implement NVENC (?), at least according to arch wiki:
> > > image
> > 
> > 
> > If I'm not mistaken, you can use NVENC as a backend for either VA-API or VDPAU, If it's VDPAU, you can use that as a backend for VA-API
> 
> Are you referring to this? https://github.com/elFarto/nvidia-vaapi-driver -- If so, this only supports decode. I would love to know how, if possible, encode can work for Chromium. Will help a friend with an Nvidia gpu a bit. This is off-topic for this PR anyway :))



> > > > > > > I'm not sure if any of you are having the same issue I am finding now, but hardware acceleration (encoding) doesn't seem to work anymore (?), aka, nvtop doesn't show encode anymore. Both myself, and a friend, are having this issue. We are both on Fedora 40, could someone else confirm this?
> > > > > > 
> > > > > > 
> > > > > > Does Linux even supports hardware encoding? I can be wrong at this point.
> > > > > 
> > > > > 
> > > > > Yes it does using vaapi (only on Intel and AMD GPUs)
> > > > 
> > > > 
> > > > According to archwiki, it's also supported on Nvidia, as NVENC supports h264, which is what vesktop uses whether or not it actually applies and works with chromium, I don't know
> > > 
> > > 
> > > I don't believe it will work on Chromium, they don't implement NVENC (?), at least according to arch wiki:
> > > image
> > 
> > 
> > If I'm not mistaken, you can use NVENC as a backend for either VA-API or VDPAU, If it's VDPAU, you can use that as a backend for VA-API
> 
> Are you referring to this? https://github.com/elFarto/nvidia-vaapi-driver -- If so, this only supports decode. I would love to know how, if possible, encode can work for Chromium. Will help a friend with an Nvidia gpu a bit. This is off-topic for this PR anyway :))

Oh, my bad. I was a bit mistaken as to how the VDPAU backend for VA-API worked lol

EmptyVisual:
to clear this up for anyone reading/not sure if it will work on their nvidia gpu, it will not. Since NVIDIA only want to support their own APIs.

Browsers use VA-API. There's two currently maintained backends for VA-API that relate to Nvidia, VAAPI > VDPAU and VAAPI > NVDEC (which is the repo linked above), both of which are decode support only, with the latter mostly being aimed at Firefox.

The only hope for Nvidia users is hopefully either open source kernel drivers nouveau or nova (the new rust kernel driver) eventually supporting VA-API encode. Unless hell freezes over and nvidia start supporting VA-API encode in their propietary driver of course. hope that clears things up :)

ILOVEPIE:
Any chance we can get #480 supported as part of this too?

PolisanTheEasyNick:
> Any chance we can get #480 supported as part of this too?

I don't know how I can test it while trying to create support

AlphaCraft9658:
> basically what the title says. streaming is now smooth on linux. finally.
> 
> TODO before merge:
> 
> * [x]  Fix resolution and framerate selection
> * [x]  Fix bug that stops users from changing the streamed window (created by allowing resolution changes)
> * [x]  Fix bug where changing windows allows you to lower the fps but not raise it
> * [ ]  Beautify modal for starting a stream(show preview and content type as a dropdown)
> * [x]  Fix issue where stream quality overlay does not update to show the current quality of the running stream
> * [x]  Enable `contentHint` and allow the user to choose between "motion" and "detail"
> * [x]  Allow users to start streams while watching another stream

I am not sure whether this is related, but on Discords or Vencord Desktop I can click on the drop-down arrow of the screen share button and adjust the resolution. Is this not possible on Vesktop, as it's based on Discord web and they don't have it there either?

ILOVEPIE:
> > Any chance we can get #480 supported as part of this too?
> 
> I don't know how I can test it while trying to create support

Just run a program in Gamescope and then run Vesktop alongside it and it will let you capture the program. for example: ``gamescope -- glxgears``.

simvux:
This works great. I'm able to stream at 40-60fps, at any resolution although at varying levels of heavy compression. 

Heavily compressed 1440p 60fps with the motion contenthint seems to be the best in general. 

However; I noticed that I get by far the best balance of compression vs framerate consistency if I first livestream something with very little moment (such as my desktop) for ~10 seconds or so, in which I'll visibly see the stream change to using way less compression. Afterwards, the rest of the stream regardless of motion, will have better results (less compression but occasional stutters and frame drops, similar to what you'd see on Windows). 
Maybe there's something with the stream settings that can be tweaked to make this seemingly improved experience be what shows right as you start the stream?

simvux:
correction: the monitor I tested on was 1080p. Me selecting 1440p was most likely just using 1080p instead.

PolisanTheEasyNick:
> This works great. I'm able to stream at 40-60fps, at any resolution although at varying levels of heavy compression.
> 
> Heavily compressed 1440p 60fps with the motion contenthint seems to be the best in general.
> 
> However; I noticed that I get by far the best balance of compression vs framerate consistency if I first livestream something with very little moment (such as my desktop) for ~10 seconds or so, in which I'll visibly see the stream change to using way less compression. Afterwards, the rest of the stream regardless of motion, will have better results (less compression but occasional stutters and frame drops, similar to what you'd see on Windows). Maybe there's something with the stream settings that can be tweaked to make this seemingly improved experience be what shows right as you start the stream?

That's because Discord is making negotiations on how to increase resolution and/or FPS to be specified in settings according to the user's bitrate. Making a stream without movement makes bitrate consumption very low (50–1000 kbit), and that gives the opportunity to increase resolution quickly.
Any ability to increase negotiation speed will be investigated in the future.

Haxorzz:
> to clear this up for anyone reading/not sure if it will work on their nvidia gpu, it will not. Since NVIDIA only want to support their own APIs.
> 
> Browsers use VA-API. There's two currently maintained backends for VA-API that relate to Nvidia, VAAPI > VDPAU and VAAPI > NVDEC (which is the repo linked above), both of which are decode support only, with the latter mostly being aimed at Firefox.
> 
> The only hope for Nvidia users is hopefully either open source kernel drivers nouveau or nova (the new rust kernel driver) eventually supporting VA-API encode. Unless hell freezes over and nvidia start supporting VA-API encode in their propietary driver of course. hope that clears things up :)

Just a question, isn't this outdated since vulkan video dropped?

regalialong:
> > to clear this up for anyone reading/not sure if it will work on their nvidia gpu, it will not. Since NVIDIA only want to support their own APIs.
> > Browsers use VA-API. There's two currently maintained backends for VA-API that relate to Nvidia, VAAPI > VDPAU and VAAPI > NVDEC (which is the repo linked above), both of which are decode support only, with the latter mostly being aimed at Firefox.
> > The only hope for Nvidia users is hopefully either open source kernel drivers nouveau or nova (the new rust kernel driver) eventually supporting VA-API encode. Unless hell freezes over and nvidia start supporting VA-API encode in their propietary driver of course. hope that clears things up :)
> 
> Just a question, isn't this outdated since vulkan video dropped?

[This Chromium issue tracks decoding](https://issues.chromium.org/issues/324003973) but isn't implemented, I'd assume by extension encoding is probably not implemented either.

PolisanTheEasyNick:
Before merge I believe we must finalize UI for picker (align sound section with stream settings section if idea is good) and remove all debug console.logs.

Kagukara:
Has anyone tested sharing a 2160p display using this rework?

PolisanTheEasyNick:
> Has anyone tested sharing a 2160p display using this rework?

There will not be enough bitrate for stable 4k sharing even on 30 FPS.

Kagukara:
> > Has anyone tested sharing a 2160p display using this rework?
> 
> There will not be enough bitrate for stable 4k sharing even on 30 FPS.

Nono, as in screensharing at 1080/1440 on a 2160p display, before it seemed I was getting worse performance than people using 1080/1440p displays. Was wondering if anyone has tried this using the rework.

PolisanTheEasyNick:
> Nono, as in screensharing at 1080/1440 on a 2160p display, before it seemed I was getting worse performance than people using 1080/1440p displays. Was wondering if anyone has tried this using the rework.

You can try add 2160 [here](https://github.com/Vencord/Vesktop/blob/main/src%2Frenderer%2Fcomponents%2FScreenSharePicker.tsx#L26) and try it by yourself if you have 4K screen

IlChitarrista:
> > > Has anyone tested sharing a 2160p display using this rework?
> > 
> > 
> > There will not be enough bitrate for stable 4k sharing even on 30 FPS.
> 
> Nono, as in screensharing at 1080/1440 on a 2160p display, before it seemed I was getting worse performance than people using 1080/1440p displays. Was wondering if anyone has tried this using the rework.

Hello, before this PR resolution setting wasn't actually working which means that, irrespective of user choice, the resolution was that of the monitor being streamed. This means that now we're sending much less data in most cases which, along with contentHint set to Motion to drop the resolution instead of framerate when bandwith isn't enough, produces a much smoother stream.

There are still some issues that need to be addressed but they seem to be somewhere in the upstream libraries.
For example, at least on GNOME/KDE Wayland, the screen sharing is worse performing than the window sharing (~60->~45-50) and, when streaming on a dual monitor setup, this becomes ever more problematic as the framerate somehow starts to heavily depend on how close the fastest updating client is to the monitor's refresh rate.

PolisanTheEasyNick:
> For example, at least on GNOME/KDE Wayland, the screen sharing is worse performing than the window sharing (~60->~45-50)
Found out that it's a Chromium upstream issue and already created [issue](https://issues.chromium.org/issues/333945842) in their bugtracker

AlphaCraft9658:
Not sure whether this is really related, but the current screen sharing implementation of the native Discord desktop client is pretty bad. Every time I stream anything on my screen, the frame rate of any currently playing game takes a noticeable hit. Is this different with this implementation?

CaptaiNiveau:
> Not sure whether this is really related, but the current screen sharing implementation of the native Discord desktop client is pretty bad. Every time I stream anything on my screen, the frame rate of any currently playing game takes a noticeable hit. Is this different with this implementation?

Yes it will be, as you're probably noticing the impact of sotware transcoding done on your CPU instead of the dedicated hardware provided by most GPUs.

regalialong:
> So this implementation will use the proper GPU encoding with H.264 or HEVC with the NVIDIA hardware transcoding module? (Also on Linux)

..... If you read up, Nvidia does not implement VA-API encoding (there is https://github.com/elFarto/nvidia-vaapi-driver which translates decoding but it doesn't do encoding), meaning there is no infrastructure to do hardware accelerated encoding for Chromium / Electron on Nvidia Linux.

AlphaCraft9658:
> > So this implementation will use the proper GPU encoding with H.264 or HEVC with the NVIDIA hardware transcoding module? (Also on Linux)
> 
> ..... If you read up, Nvidia does not implement VA-API encoding (there is https://github.com/elFarto/nvidia-vaapi-driver which translates decoding but it doesn't do encoding), meaning there is no infrastructure to do hardware accelerated encoding for Chromium / Electron on Nvidia Linux.

OBS studio can use hardware encoding, so I don't really see why a Discord client would be able to do so as well. Even if it wouldn't use VA-API, there should still be a way.

kaitlynkittyy:
> > > So this implementation will use the proper GPU encoding with H.264 or HEVC with the NVIDIA hardware transcoding module? (Also on Linux)
> > 
> > 
> > ..... If you read up, Nvidia does not implement VA-API encoding (there is https://github.com/elFarto/nvidia-vaapi-driver which translates decoding but it doesn't do encoding), meaning there is no infrastructure to do hardware accelerated encoding for Chromium / Electron on Nvidia Linux.
> 
> OBS studio can use hardware encoding, so I don't really see why a Discord client would be able to do so as well. Even if it wouldn't use VA-API, there should still be a way.

That's not our problem to solve, as that's an issue to bring up with upstream chromium. 
image
According to the archwiki, chromium only supports VA-API based hardware encoding, whereas something using FFMPEG or GStreamer as a backend would support NVENC.

AlphaCraft9658:
> > > > So this implementation will use the proper GPU encoding with H.264 or HEVC with the NVIDIA hardware transcoding module? (Also on Linux)
> > > 
> > > 
> > > ..... If you read up, Nvidia does not implement VA-API encoding (there is https://github.com/elFarto/nvidia-vaapi-driver which translates decoding but it doesn't do encoding), meaning there is no infrastructure to do hardware accelerated encoding for Chromium / Electron on Nvidia Linux.
> > 
> > OBS studio can use hardware encoding, so I don't really see why a Discord client would be able to do so as well. Even if it wouldn't use VA-API, there should still be a way.
> 
> That's not our problem to solve, as that's an issue to bring up with upstream chromium. 
> image
> According to the archwiki, chromium only supports VA-API based hardware encoding, whereas something using FFMPEG or GStreamer as a backend would support NVENC.

So either Chromium could implement it, which I don't really think will happen, but I might be wrong, or Vesktop could implement an additional backend like ffmpeg, gstreamer or similar, which may also be rather complicated and maybe not even possible.

AlphaCraft9658:
> > > > So this implementation will use the proper GPU encoding with H.264 or HEVC with the NVIDIA hardware transcoding module? (Also on Linux)
> > > 
> > > 
> > > ..... If you read up, Nvidia does not implement VA-API encoding (there is https://github.com/elFarto/nvidia-vaapi-driver which translates decoding but it doesn't do encoding), meaning there is no infrastructure to do hardware accelerated encoding for Chromium / Electron on Nvidia Linux.
> > 
> > OBS studio can use hardware encoding, so I don't really see why a Discord client would be able to do so as well. Even if it wouldn't use VA-API, there should still be a way.
> 
> That's not our problem to solve, as that's an issue to bring up with upstream chromium. 
> image
> According to the archwiki, chromium only supports VA-API based hardware encoding, whereas something using FFMPEG or GStreamer as a backend would support NVENC.

So either Chromium could implement it, which I don't really think will happen, but I might be wrong, or Vesktop could implement an additional backend like ffmpeg, gstreamer or similar, which may also be rather complicated and maybe not even possible.

AlphaCraft9658:
> > > > So this implementation will use the proper GPU encoding with H.264 or HEVC with the NVIDIA hardware transcoding module? (Also on Linux)
> > > 
> > > 
> > > ..... If you read up, Nvidia does not implement VA-API encoding (there is https://github.com/elFarto/nvidia-vaapi-driver which translates decoding but it doesn't do encoding), meaning there is no infrastructure to do hardware accelerated encoding for Chromium / Electron on Nvidia Linux.
> > 
> > OBS studio can use hardware encoding, so I don't really see why a Discord client would be able to do so as well. Even if it wouldn't use VA-API, there should still be a way.
> 
> That's not our problem to solve, as that's an issue to bring up with upstream chromium. 
> image
> According to the archwiki, chromium only supports VA-API based hardware encoding, whereas something using FFMPEG or GStreamer as a backend would support NVENC.

So either Chromium could implement NVENC encoding and decoding support, which I don't really think will happen, but I might be wrong, or Vesktop could implement an additional backend like ffmpeg, gstreamer or similar, which may also be rather complicated and maybe not even possible. Right?

AlphaCraft9658:
Somehow the post was displayed twice for me. Deleted it, but both versions were gone. Reposted it. Sorry.

kaitlynkittyy:
> > > > > So this implementation will use the proper GPU encoding with H.264 or HEVC with the NVIDIA hardware transcoding module? (Also on Linux)
> > > > 
> > > > 
> > > > ..... If you read up, Nvidia does not implement VA-API encoding (there is https://github.com/elFarto/nvidia-vaapi-driver which translates decoding but it doesn't do encoding), meaning there is no infrastructure to do hardware accelerated encoding for Chromium / Electron on Nvidia Linux.
> > > 
> > > 
> > > OBS studio can use hardware encoding, so I don't really see why a Discord client would be able to do so as well. Even if it wouldn't use VA-API, there should still be a way.
> > 
> > 
> > That's not our problem to solve, as that's an issue to bring up with upstream chromium.
> > image
> > According to the archwiki, chromium only supports VA-API based hardware encoding, whereas something using FFMPEG or GStreamer as a backend would support NVENC.
> 
> So either Chromium could implement NVENC encoding and decoding support, which I don't really think will happen, but I might be wrong, or Vesktop could implement an additional backend like ffmpeg, gstreamer or similar, which may also be rather complicated and maybe not even possible. Right?

Pretty much, yeah. We couldn't add another backend because vesktop is based on electron, which uses chromium. adding anything like that would be a monumental undertaking.

AlphaCraft9658:
> > > > > > So this implementation will use the proper GPU encoding with H.264 or HEVC with the NVIDIA hardware transcoding module? (Also on Linux)
> > > > > 
> > > > > 
> > > > > ..... If you read up, Nvidia does not implement VA-API encoding (there is https://github.com/elFarto/nvidia-vaapi-driver which translates decoding but it doesn't do encoding), meaning there is no infrastructure to do hardware accelerated encoding for Chromium / Electron on Nvidia Linux.
> > > > 
> > > > 
> > > > OBS studio can use hardware encoding, so I don't really see why a Discord client would be able to do so as well. Even if it wouldn't use VA-API, there should still be a way.
> > > 
> > > 
> > > That's not our problem to solve, as that's an issue to bring up with upstream chromium.
> > > image
> > > According to the archwiki, chromium only supports VA-API based hardware encoding, whereas something using FFMPEG or GStreamer as a backend would support NVENC.
> > 
> > So either Chromium could implement NVENC encoding and decoding support, which I don't really think will happen, but I might be wrong, or Vesktop could implement an additional backend like ffmpeg, gstreamer or similar, which may also be rather complicated and maybe not even possible. Right?
> 
> Pretty much, yeah. We couldn't add another backend because vesktop is based on electron, which uses chromium. adding anything like that would be a monumental undertaking. (and most likely impossible)

So In conclusion, either Chromium or Electron need to implement NVENC support or we need a VAAPI encoding implementation for the proprietary NVIDIA drivers? And both are not directly related to the development of Vesktop.

kaitlynkittyy:
yep

413120:
Sorry if it has gotten buried in here somewhere, but is there a solution for the stream lagging/skipping based on mouse movement over/in the streamed window? I believe I may have seen a comment about it being specific to certain DEs? I'm using kde plasma in wayland.

PolisanTheEasyNick:
> Sorry if it has gotten buried in here somewhere, but is there a solution for the stream lagging/skipping based on mouse movement over/in the streamed window? I believe I may have seen a comment about it being specific to certain DEs? I'm using kde plasma in wayland.

We believe this is a Chromium upstream [issue](https://issues.chromium.org/issues/333945842), because this behavior is not reproducing in Firefox browser.
For temp solution, you can try to stream application window instead of whole screen. As far as I remember, this issue is not reproducible for windows because the stream will be updated only when window is updated (and will receive stable 60fps on 60fps updating window).

Rune580:
> > Sorry if it has gotten buried in here somewhere, but is there a solution for the stream lagging/skipping based on mouse movement over/in the streamed window? I believe I may have seen a comment about it being specific to certain DEs? I'm using kde plasma in wayland.
> 
> We believe this is a Chromium upstream [issue](https://issues.chromium.org/issues/333945842), because this behavior is not reproducing in Firefox browser. For temp solution, you can try to stream application window instead of whole screen. As far as I remember, this issue is not reproducible for windows because the stream will be updated only when window is updated (and will receive stable 60fps on 60fps updating window).

this issue can be re-produced on firefox using vdo.ninja, I'm also running KDE wayland and have the same problem.

PolisanTheEasyNick:
> this issue can be re-produced on firefox using vdo.ninja, I'm also running KDE wayland and have the same problem.

Thank you for useful info. Anyway, that means that mouse issue is some upstream issue and we in Vesktop code can't do anything about it :c

AlphaCraft9658:
> yep

Btw, which codec is used for Discord screen sharing? VP9?

kaitlynkittyy:
> > yep
> 
> Btw, which codec is used for Discord screen sharing? VP9?

H264, as of right now.
I may look into H265 at some point, but that's not likely to happen.

regalialong:
I don't mean to be rude but I'm subscribed to this PR because I'd like to track the development of it, I would prefer if exploratory questions like that could be moved to IM.

There's a lot of noise here from lots of people waiting for this PR which drowns out useful information (which is why iirc we talked 3 times about Nvidia encode despite it not happening).

kaitlynkittyy:
> I don't mean to be rude but I'm subscribed to this PR because I'd like to track the development of it, I would prefer if exploratory questions like that could be moved to IM.
> 
> There's a lot of noise here from lots of people waiting for this PR which drowns out useful information (which is why iirc we talked 3 times about Nvidia encode despite it not happening).

That's a fair point. Further questions can be directed at my discord: kaitlynyaa (although this PR is due for a merge really soon tbf)

EndenDragon:
Any support for soundsharing on MacOS?

kaitlynkittyy:
> Any support for soundsharing on MacOS?
> It would be nice to be able to pick from the list of Mic-in inputs to share.

not in this pr, no. I can look into it now that I have a macos device to test on, but I cannot confirm anything, as I have absolutely no idea how that would be done.

lewisakura:
Can you build this PR as a Flatpak (https://github.com/flathub/dev.vencord.Vesktop) and validate it works? We have a decent amount of users using it via Flatpak now so it needs to be tested

kaitlynkittyy:
> Can you build this PR as a Flatpak (https://github.com/flathub/dev.vencord.Vesktop) and validate it works? We have a decent amount of users using it via Flatpak now so it needs to be tested

I also use it as a flatpak, and I just confirmed that I am able to build and run it as a flatpak, with the streaming modifications. I didn't do any sort of rigorous testing, but I did confirm that it works.

src/main/index.ts
```diff
@@ -27,14 +27,18 @@ process.env.VENCORD_USER_DATA_DIR = DATA_DIR;
 function init() {
     const { disableSmoothScroll, hardwareAcceleration } = Settings.store;
 
-    if (hardwareAcceleration === false) app.disableHardwareAcceleration();
+    if (hardwareAcceleration === false) {
+        app.disableHardwareAcceleration();
+    } else {
+        app.commandLine.appendSwitch("enable-features", "VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,VaapiVideoDecoder");
```
andrepapoti, line 33: `enable-features` doesn't seen to be a [supported cli flag](https://www.electronjs.org/docs/latest/api/command-line-switches)

rurigk: https://chromium.googlesource.com/chromium/src/+/main/docs/configuration.md#Features

It's a chromium flag not a electron flag

src/main/index.ts
```diff
@@ -27,14 +27,18 @@ process.env.VENCORD_USER_DATA_DIR = DATA_DIR;
 function init() {
     const { disableSmoothScroll, hardwareAcceleration } = Settings.store;
 
-    if (hardwareAcceleration === false) app.disableHardwareAcceleration();
+    if (hardwareAcceleration === false) {
+        app.disableHardwareAcceleration();
+    } else {
```
Vendicated, line 32: probably best to only do this on linux?
```suggestion
    } else if (process.platform === "linux") {
```

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -380,35 +444,83 @@ function ModalComponent({
...
-                                    width,
-                                    height
-                                }
-                            });
+                        try {
+                            var conn = [...MediaEngineStore.getMediaEngine().connections].find(
+                                connection => connection.streamUserId === UserStore.getCurrentUser().id
+                            );
+                        } catch {
+                            console.log("No current stream.");
```
Vendicated, line 457: this log should be removed

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -380,35 +444,83 @@ function ModalComponent({
...
+                                const newConstraints = {
+                                    ...constraints,
+                                    frameRate,
+                                    width: { min: 640, ideal: width, max: 2560 },
+                                    height: { min: 480, ideal: height, max: 1440 },
+                                    advanced: [{ width: width, height: height }],
+                                    resizeMode: "none"
+                                };
+                                track.applyConstraints(newConstraints).then(() => {
+                                    console.log("Applied constraints from ScreenSharePicker successfully.");
```
Vendicated, line 475: either remove these logs or use a Logger to be more informative

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -380,35 +444,83 @@ function ModalComponent({
...
+                                    console.log("New constraints:", track.getConstraints());
+                                });
+
+                                // changing stream quality description
```
Vendicated, line 479: don't explain self explanatory things via comments
```suggestion
```

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -380,35 +444,83 @@ function ModalComponent({
...
+                                conn.videoStreamParameters[0].maxFrameRate = Number(settings.fps);
+                                conn.videoStreamParameters[0].maxResolution.height = Number(settings.resolution);
+                                conn.videoStreamParameters[0].maxResolution.width = Math.round(height * (16 / 9));
+                            }
+                        } catch {
```
Vendicated, line 484: this try catch seems duplicated, and the error message misleading

merge them and add more logical error handling, or remove them altogether

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -380,35 +444,83 @@ function ModalComponent({
...
+                            });
 
-                        submit({
-                            id: selected!,
-                            ...settings
-                        });
+                            // reapply contraints after some time to let discord resubmit stream
+                            // i believe there MUST be way to do it cleaner..
+                            if (conn) {
+                                setTimeout(() => {
```
Vendicated, line 496: avoid weird hacks like this. why is this necessary in the first place?

src/renderer/components/screenSharePicker.css
```diff
@@ -128,3 +128,10 @@
 .vcd-screen-picker-audio {
     margin-bottom: 0;
 }
+
+.vcd-screen-picker-hint-description {
+    color: var(--header-secondary);
+    font-size: 14px;
+    line-height: 20px;
```
Vendicated, line 135: why? doesn't make a lot of sense

prefer paddings/margins to do extra spacing

src/renderer/patches/screenShareFixes.ts
```diff
@@ -23,6 +24,27 @@ if (isLinux) {
...
+            frameRate,
+            width: { min: 640, ideal: width, max: 2560 },
+            height: { min: 480, ideal: height, max: 1440 },
+            advanced: [{ width: width, height: height }],
+            resizeMode: "none"
+        };
+
+        track.applyConstraints(newConstraints).then(() => {
+            console.log("Applied constraints from ScreenShareFixes successfully.");
+            console.log("New constraints:", track.getConstraints());
```
Vendicated, line 45: same as above

Vendicated (changes requested):

Vendicated:
thanks all for working on this!!

src/renderer/components/ScreenSharePicker.tsx
PolisanTheEasyNick: About `setTimeout`: we use it because after `submit()` function work (even using `await` with `then`) discord take some time to set their default screenres x 30 fps settings. If we change constraints without timeout, discord just will set up 30 fps for it instead of picked user settings
I really wish we could do this without `setTimeout()` :c

Vendicated (approved):

PolisanTheEasyNick:
Thank you all for work and help ^-^

BuonHobo:
When will this reach the flathub version?

SpyTec:
> When will this reach the flathub version?

When a new release is pushed for Vesktop itself, which needs to happen first, you can check the updates on the flathub repo

https://github.com/flathub/dev.vencord.Vesktop

Vendicated:
expect a release some time in the next few days. still want to land a few things in vesktop before making it, and maybe experiment with electron 30

TheRsKing:
How do i build the flatpak from source

SpyTec:
> How do i build the flatpak from source

@TheRsKing https://github.com/Vencord/Vesktop?tab=readme-ov-file#building-from-source

PR #514, opened 2024-04-17
Update packages

cloudcloudskies:
My bad if there's code needs to be ported. I got 0 knowledge I just updated the packages so don't blame me.


PolisanTheEasyNick:
30 Electron does not work with current Vesktop version :c

Vendicated:
but why

this just seems like contribution farming

Issue #526, opened 2024-04-20
--enable-features flags not passed into Electron


takase1121:
**Describe the bug**

I am trying to get VAAPI Video Decoding to work on amdgpu, which needs a bunch of flags and some amount of prayers.
Since Vesktop tries to append it's own `enable-features` flags this caused it to override the previous (user added) flags. I have no proof of this happening, but in my testing removing this line made it work:

https://github.com/Vencord/Vesktop/blob/8eaa5206b98ce7029b6879cc2ce361c2ae5f6239/src/main/index.ts#L33

Those flags has good intentions, but it will take more flags (to force Electron to use ANGLE and ANGLE to use Vulkan) to make it work, and it's not possible to set them if it got overridden by the flags above.

**To Reproduce**

**Not working**

1. Install `vesktop-git` from AUR.
2. run `vesktop --use-gl=angle --use-angle=vulkan --ignore-gpu-blocklist --enable-gpu-rasterization --enable-zero-copy --enable-features=Vulkan,VulkanFromANGLE,DefaultANGLEVulkan,VaapiIgnoreDriverChecks,VaapiVideoDecoder,PlatformHEVCDecoderSupport,UseMultiPlaneFormatForHardwareVideo`
3. womp womp womp womp

**Making it work**

1. clone Vesktop, remove the line in https://github.com/Vencord/Vesktop/blob/8eaa5206b98ce7029b6879cc2ce361c2ae5f6239/src/main/index.ts#L33
2. `pnpm i && pnpm build`
3. `electron29 --use-gl=angle --use-angle=vulkan --ignore-gpu-blocklist --enable-gpu-rasterization --enable-zero-copy --enable-features=Vulkan,VulkanFromANGLE,DefaultANGLEVulkan,VaapiIgnoreDriverChecks,VaapiVideoDecoder,PlatformHEVCDecoderSupport,UseMultiPlaneFormatForHardwareVideo .`
4. Electron tries to use GPU to decode (good effort but still terrible)

**Expected behavior**

Vesktop can detect if `--enable-features` is set, get all the flags and append its own flags. After that, Vesktop will append it back to `app.cmdLine`.

**Screenshots**

**Working**
![image](https://github.com/Vencord/Vesktop/assets/20792268/7ae08ff8-154a-4fbe-96e7-88e50ccbe488)

**Not Working**
[...]

PR #527, opened 2024-04-20
Allow users to pass --enable/disable-features flag

takase1121:
This PR implements the solution in #526.

I am also looking to adding the flags to Vesktop, unfortunately it is impossible to set the argument in time as `app.getGPUInfo()` is async and I don't know if Electron has already initialized the GUI backend by then. It might be worth looking into as a toggle in the settings (probably something like "Experimental Video Decode Acceleration on AMD" but this is so specific it's hard to come out with a suitable name).

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -27,10 +27,12 @@ process.env.VENCORD_USER_DATA_DIR = DATA_DIR;
 function init() {
     const { disableSmoothScroll, hardwareAcceleration } = Settings.store;
 
+    const enableFeatures: string[] = [];
+
     if (hardwareAcceleration === false) {
         app.disableHardwareAcceleration();
     } else {
-        app.commandLine.appendSwitch("enable-features", "VaapiVideoDecodeLinuxGL,VaapiVideoEncoder,VaapiVideoDecoder");
+        enableFeatures.push("VaapiVideoDecodeLinuxGL", "VaapiVideoEncoder", "VaapiVideoDecoder");
     }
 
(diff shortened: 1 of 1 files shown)


Vendicated:
this is meh. if the user wants to override one switch, it will disable everything else

why not merge our switches into the user switches? could do the same for disable-features

```js
const features = app.commandLine.getSwitchValue("enable-features");
features.push("Feature1", "Feature2");

app.commandLine.appendSwitch("enable-features", [...new Set(features)].filter(Boolean).join(","));
```

takase1121:
> this is meh. if the user wants to override one switch, it will disable everything else
> 
> why not merge our switches into the user switches? could do the same for disable-features
> 
> ```js
> const features = app.commandLine.getSwitchValue("enable-features").split(",");
> features.push("Feature1", "Feature2");
> 
> app.commandLine.appendSwitch("enable-features", [...new Set(features)].filter(Boolean).join(","));
> ```

That would be an option. I'm just mainly concerned with conflicting switches.m, but I think it will probably be fine.

takase1121:
> I'm just mainly concerned with conflicting switches.m, but I think it will probably be fine.

This might not be an issue at all. When a flag appears in both `enable-features` and `disable-features`, it will be disabled. If someone wants to override Vesktop's flags, they can just disable them from the CLI. Assuming Vesktop doesn't disable something crazy by default (right not it doesn't) this will work well.

P.S.: Is this something worth documenting?

Vendicated:
thank you!

PR #533, opened 2024-04-22
mac: add entitlements needed for camera/microphone access

ssalggnikool:
## Problem
With [Vesktop v1.5.1](https://github.com/Vencord/Vesktop/releases/tag/v1.5.1) there was a problem with Vesktop not requesting for permission and rather denying access immediately, I assume it is due to it no longer having an ad-hoc signature, though I don't exactly know the specifics but.... **adding/specifying the entitlements manually** seems to work fine.

Was pretty simple to fix, just adding the entitlements file to `build/` and electron-builder will sign with those entitlements. (not sure why it doesn't use the ones provided in `package.json`..) 

### Before
image
image

### After 
image

image

## Note
Also, assuming that it required proper entitlements when signed; camera access and access to bluetooth devices most likely didn't exactly work as well.. Though correct me if I'm wrong, I'm not gonna test it as I don't have devices neccessary.

I have also tested this on an M2 mac with a developer certificate.

### DIFF
--- a/build/entitlements.mac.plist
+++ b/build/entitlements.mac.plist
@@ -1,0 +1,21 @@
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+	<key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+	<true/>
+	<key>com.apple.security.cs.allow-jit</key>
+	<true/>
+	<key>com.apple.security.network.client</key>
+	<true/>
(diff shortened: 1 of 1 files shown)


ssalggnikool:
Also when testing permissions for the microphone, you can get it to prompt back up if you've already allowed it with this command: 
```sh
sqlite3 ~/Library/Application\ Support/com.apple.TCC/TCC.db`
delete from access where service='kTCCServiceMicrophone' and client='dev.vencord.vesktop';```

Though would not recommend you keep vencord opened when doing so because it will prompt you that it doesn't have access when it indeed does. So either kill the application before doing it or kill it after.

For screen recording it should ask for permissions and not require any new entitlements, so allowing that in settings with the prompt wouldn't need anything fixing within vencord

PR #546, opened 2024-04-25
Fix window listeners bug

zt64:


src/main/mainWindow.ts
```diff
@@ -409,6 +409,8 @@ function createMainWindow() {
     }));
     win.setMenuBarVisibility(false);
 
+    win.on("ready-to-show", () => initWindowBoundsListeners(win));
```
Vendicated, line 412: try moving this into the `mainWin.webContents.on("did-finish-load"`

Vendicated:
not needed anymore right?

PR #563, opened 2024-05-02
move patch to vencord core plugin, DO NOT MERGE WITHOUT MERGING MY OTHER PR INTO VENCORD FIRST

kaitlynkittyy:
just shuffling a patch around to make sure it's enabled for chromium users too :3

### DIFF
--- a/src/renderer/components/ScreenSharePicker.tsx
+++ b/src/renderer/components/ScreenSharePicker.tsx
@@ -63,20 +63,6 @@ addPatch({
                 match: /this.localWant=/,
                 replace: "$self.patchStreamQuality(this);$&"
             }
-        },
-        {
-            find: "x-google-max-bitrate",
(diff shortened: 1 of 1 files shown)


powerofthe69:
excuse the bump but I was hoping to know when we might expect this change to be merged into Vencord and Vesktop? The current bitrate cap makes my streams sad, especially on more complex games :(

kaitlynkittyy:
It's already a part of vesktop, but for other web clients (chromium), it is ready to merge afaik.

kaitlynkittyy:
To clarify, I'm waiting for the plugin to merge into vencord before this is ready to be removed from vesktop. As of right now the bitrate cap is removed on vesktop and not web, but once it's added to vencord, it can be freely removed from vesktop, as vencord will make the same patch.

PR #565, opened 2024-05-02
macOS: Add customized dmg background

ssalggnikool:
image
# 
Color pallete based on lightmode varient of https://vencord.dev

###### Suggestions welcome!


Curve:
Looks great!

ssalggnikool:
FYI, darkmode isn't exactly an option here as the text will not adopt to the background color unless I do some weird shinanigans with the image

package.json
```diff
@@ -131,6 +131,26 @@
                 "com.apple.security.device.camera": true
             }
         },
+        "dmg": {
+            "background": "assets/background.tiff",
```
Vendicated, line 135: is this included at build time? if yes it should be in the build folder, not assets

ssalggnikool: The dmg iirc is created after vesktop is built

Vendicated: try moving it into build and see if it still works

ssalggnikool: Done, and yes it compiles fine.

Vendicated: yep i figured! assets/ is for runtime assets, build/ for compile time only stuff

Vendicated:
looks cute! 

> FYI, darkmode isn't exactly an option here as the text will not adopt to the background color unless I do some weird shinanigans with the image

by text, do you mean the "Vesktop.app"?  how do other apps usually handle light/dark mode?

package.json
ssalggnikool: I didn't even know that tbf, I just thought it would be appropriate to have it be in another directory

ssalggnikool:
> looks cute!
> 
> > FYI, darkmode isn't exactly an option here as the text will not adopt to the background color unless I do some weird shinanigans with the image
> 
> by text, do you mean the "Vesktop.app"? how do other apps usually handle light/dark mode?

Yes I do mean the titles of the files, FYI a dmg is just a glorified iso file that adds itself as a location in the default macos file manager named 'Finder', this is used to install applications on macOS (such as Vesktop/Discord)

The background here is just an image file, in some custom directory window

> how do other apps usually handle light/dark mode?

They kinda don't, they either use the macOS appstore and not rely on a dmg, Or, either use a custom background, or no background at all.

Examples: 

image

###### They add a white box below where the title is gonna be, since it cannot be something other than black with a background.

PR #568, opened 2024-05-02
fix: Screenshare UI for non-linux systems

PolisanTheEasyNick:
Now on Windows system screenshare popup will look like this:
image
And on Linux it is like before:
image

~~Also, it must be fixed now in MacOS systems too, but I need someone to test it.~~ it works on MacOS too.


Golonchy:
works, epic

Vendicated:
thank you!

PR #517, opened 2024-04-18
feat: Tray icon voice detection

PolisanTheEasyNick:
epic pr🚀 cooking explodin
Resolves #294, partially resolves #667
- [x] Show icons depending on current voice state

Any ideas about improving PR structure are welcome ^^

---

### Voice Detection Icons
image image image image

---

### Notification Badge on top of Tray Icon
image

Special thanks to @MrGarlic1 and @Covkie 

Also, on my repo, there is branch `tray-icon-old` where all changes with tray icons customizations, colors, and so on

PR #576, opened 2024-05-04
feat: Allow tray icon customization

### LINKED ISSUE #571: Add a new option to change the tray icon used by Vesktop
Hi there. I would like to have a command-line flag or GUI option to change the icon used by Vesktop on the system tray. For the GUI, it could be something together with the "Tray Icon" option in "Vesktop Setttings".

My reasoning for this request is because I personally prefer to use original Discord icons.
To clarify, I'm not requesting to add Discord icons inside Vesktop itself. What I'm requesting is to let the user choose custom icons from their PC directly.

For the app itself, I can do it using the `.desktop` file on my DE (KDE Plasma), but I can't modify the VC icon due to it being hardcoded in Vesktop itself. (I could, but I would have to recompile the program, and that's not very convenient when updating.)

Context of this issue: see issues #448 and #570.

One thing that I'm not sure is about having `static.ico` and `static.png`. If I could modify Vesktop's tray icon, which extension would I have to use? Or would I need to have both file extensions for two copies of the same image? I think it would be interesting to use only `.ico` or `.png` to make it simpler (if needed).

Thanks in advance to all contributors, and a special thanks to Vendicated for having a very friendly behavior.

image

MrGarlic1:
Added option in Vesktop settings to select a custom tray icon, resolves  #571 

image

Only tested on Fedora 40 KDE. May need additional testing on Windows, I don't currently have access to a Windows device right now.


Vendicated:
wow looks really good so far especially for a first contribution! 

i'm not sure about the ui though, it would probably be much better to actually show the image that is chosen and maybe have it so you can click the image to change it?

also i can see a bit of inconsistent formatting, make sure you have prettier properly configured!

MrGarlic1:
I had some issues with displaying an image from the local filesystem on the Vesktop settings page. I moved the tray icon setting to a tray icon context menu item, where I think it makes more sense to be. (old method kept on a branch in case this ends up being worse)

image

When selected, it brings up this window which has the same functionality as the setting did before while also showing the current icon. This could probably use some additional formatting to make it look better. 

image

MrGarlic1:
> Many thanks for your PR!
> 
> To be honest, I don't think an option as "important" as this should be so easily accessible with via right-click. Vendicated's idea ends up fitting better (of displaying the icon on the settings page next to the "Tray Icon" option, instead of your idea of adding new buttons).

That's my end goal too, I just shied away due to literally never having used css or html. After much learning (read: pain), the recent commits moved it back to being in the Vesktop settings menu, with a better UI:
 
image

When hovered (test icon shown):

image

Currently the edit icon image is a hard link to a discord attachment which I don't think are permanent anymore. Unsure if I should store the icon somewhere in the project itself or try and find a permanent URL for it.

src/renderer/components/settings/TrayIconImagePicker.tsx
```diff
@@ -0,0 +1,73 @@
...
+                            className="tray-icon-image"
+                            src={VesktopNative.tray.getTrayIcon()}
+                            alt="hello"
+                            width="48"
+                            height="48"
+                        ></img>
+                        <input
+                            className="edit-button"
+                            type="image"
+                            src="https://cdn.discordapp.com/attachments/895550066453012480/1236925384482619433/I1oxwE7.png?ex=6639c808&is=66387688&hm=5c5f47a0c06b2e0580fd5494386bfbeebc001c20e3fd64f26783f360b12162ed&"
```
Vendicated, line 51: this cdn link will expire and break

src/renderer/components/settings/settings.css
```diff
@@ -11,4 +11,62 @@
 
 .vcd-settings-title {
     margin-bottom: 0.5rem;
-}
\ No newline at end of file
+}
+
+.tray-icon-wrap {
```
Vendicated, line 16: all classes should be prefixed
```suggestion
.vcd-tray-icon-wrap {
```

src/renderer/components/settings/TrayIconImagePicker.tsx
Vendicated: oh you said this already! well it's a bit tough to make it permanent, you could upload it as a file and use the github raw link

src/renderer/components/settings/settings.css
```diff
@@ -11,4 +11,62 @@
...
+    transition: 0.3s ease;
+    visibility: visible;
+    opacity: 1;
+}
+
+#tray-setting {
+    height: 48px;
+    position: relative;
+}
+#colLeft {
```
Vendicated, line 53: this css is really cursed

look into css grid/flex layout! but i can also help with the ui if you'd like

MrGarlic1: Thanks for the tip! I changed to a flex layout but this is definitely not my strong suit so feel free to offer some suggestions

src/renderer/components/settings/TrayIconImagePicker.tsx
MrGarlic1: Changed to a github link as suggested. It is linked to Vencord/Vesktop so it will show correctly if merged

src/renderer/components/settings/TrayIconImagePicker.tsx
```diff
@@ -0,0 +1,73 @@
...
+                        className="vcd-tray-icon-image"
+                        src={VesktopNative.tray.getTrayIcon()}
+                        alt="hello"
+                        width="48"
+                        height="48"
+                    ></img>
+                    <input
+                        className="vcd-edit-button"
+                        type="image"
+                        src="https://raw.githubusercontent.com/Vencord/Vesktop/main/static/pencil-edit-icon.svg"
```
Cynosphere, line 50: Why not just use the pencil icon provided by the client wrapped in a Clickable? It can be grabbed via `findByCode(".replaceIcon", ".PencilIcon")`

MrGarlic1: Thanks for the tip! I am unable to import findByCode from @vencord/types/webpack, and I'm unsure why. I have tried with both my fork and the main branch here.

MrGarlic1: Attempted this, `src={findByCode(".replaceIcon", ".PencilIcon")}` resulted in no icon being found. May be a case of user error, though.

Cynosphere: > Attempted this, `src={findByCode(".replaceIcon", ".PencilIcon")}` resulted in no icon being found. May be a case of user error, though.

It is its own component, not a URL.

MrGarlic1: I may need some help making this change then. Most things I try end up breaking, and I am unsure of where to start here.

Thanks again, and sorry about this.

TwilightKiddy:
Played around with it under Arch with KDE Plasma, working mighty fine with the exception of a missing edit icon.
Is it possible to allow `.svg` and `.ico` files to be chosen? They are far more common for distributing app icons, as far as I'm aware.

MrGarlic1:
> Played around with it under Arch with KDE Plasma, working mighty fine with the exception of a missing edit icon. Is it possible to allow `.svg` and `.ico` files to be chosen? They are far more common for distributing app icons, as far as I'm aware.

I tried doing this earlier, but Electron's NativeImage (which Tray uses) can only create images from .png and .jpg files. (RIP GIF tray icons)

rootiest:
> To the author of this PR: could you test if this also changes the icon when pressing alt+tab? (In KDE, there is the Vencord icon when you press alt+tab.)

On KDE I can override the icon used in the task switcher by:

- First change the icon for Vesktop in your app menu to your preferred icon, then
- Right-click on the Vesktop titlebar and go to More Actions > Configure Special Application Settings
- Click the Detect Window Properties button
- Select the Vesktop window with the crosshair
- Click on the first one: Window Class Application
- Click Add Property then Desktop Filename
- Make sure this matches the name of the .desktop file used in your App menu, typically just "vesktop"
- Choose "Apply Now"
- Click Apply, and then verify the icon has changed to your liking
- Change "Apply Now" to "Force"
- Save the new rule.
- Close and reopen Vesktop to verify the icon change sticks.
- Enjoy your new icon in the App menu, taskbar, and app switcher (not tray, yet!)

src/renderer/components/settings/TrayIconImagePicker.tsx
Huderon: You can just grab it like `const { PencilIcon } = findByPropsLazy("PencilIcon");` and then use it like this, I think 

```js
<PencilIcon
    className="vcd-edit-button"
    width="40"
    height="40"
    onClick={async () => {
        const choice = await VesktopNative.fileManager.selectTrayIcon();
        switch (choice) {
            case "cancelled":
                return;
            case "invalid":
                Toasts.show({
                    message: "Please select a valid .png or .jpg image!",
                    id: Toasts.genId(),
                    type: Toasts.Type.FAILURE
                });
                return;
        }
        settings.trayIconPath = choice;
    }}
/>
```

MrGarlic1: Thanks! It was the findbyprops that I was missing.

SomeAspy:
It might be worth adding an option to use the favicon the discord website sends, because this dynamically updates the number of pings for just about any number. We do this in armcord, if you would like to see a working example.
https://github.com/ArmCord/ArmCord/blob/573c96b4ba80705b1bc5f85c5a737258481023e9/src/discord/window.ts#L168

marek2001:
Someone pleas push this. I would like to change tray icon. And that would be perfect.

imshinyu:
any updates?

seraisonfire:
Update?

TwilightKiddy:
> Update?

I'm not sure it will ever get merged, considering #517 exists.
If you want this feature right here and now, you can check out [GoofCord](https://github.com/Milkshiift/GoofCord). It supports Vencord plugins and allows changing the tray icon.

Kyriella:
Discord crashes and force re-loads every time I try to open the Vesktop Settings tab with this PR checked out. I'm on Arch Linux using Xfce4. Let me know if you need any other information!

simplestan1:
can we please merge this 1 year and 6 month old pr??? or close it??

Vendicated:
Superseded by #1179 but thank you anyway

PR #589, opened 2024-05-08
Finally added the submenu for changing the splash animation

sz6084:
Added a sub-menu for custom splash animation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,10 @@ If you don't know the difference, pick the Installer.
 
 ### Mac
 
-If you don't know the difference, pick amd64
+If you don't know the difference, pick the Intel build.
 
-- [amd64 / x86_64](https://vencord.dev/download/vesktop/amd64/dmg)
-- [arm64 / aarch64](https://vencord.dev/download/vesktop/arm64/dmg)
+- [Intel build (amd64)](https://vencord.dev/download/vesktop/amd64/dmg)
+- [Apple Silicon (arm64)](https://vencord.dev/download/vesktop/arm64/dmg)
 
 ### Linux
 
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
         "typescript": "^5.4.2",
         "xml-formatter": "^3.6.2"
     },
-    "packageManager": "pnpm@8.11.0",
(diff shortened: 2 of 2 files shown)


Vendicated:
please do not change unrelated files

src/renderer/components/settings/CustomSplashAnimation.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { SettingsComponent } from "./Settings";
+
+export const CustomSplashAnimation: SettingsComponent = ({ settings }) => {
+    return (
+        <>
+        
+        <Forms.FormText>
+            The animation on the splash window is loaded from{" "}
+            {settings.splashAnimationPath ? (
+                <a
```
Cynosphere, line 18: Use `Clickable` instead of an anchor tag.

src/renderer/components/settings/VencordLocationPicker.tsx
```diff
@@ -11,6 +11,45 @@ import { SettingsComponent } from "./Settings";
 export const VencordLocationPicker: SettingsComponent = ({ settings }) => {
     return (
         <>
+        {/*
```
Cynosphere, line 14: Remove commented out code

src/main/index.ts
```diff
@@ -74,6 +74,10 @@ function init() {
 
         registerScreenShareHandler();
         registerMediaPermissionsHandler();
+        //register file handler so we can load the custom splash animation from the user's filesystem
+        protocol.handle("splash-animation", () => {
+            return net.fetch("file:///"+splashAnimationPath);
```
Cynosphere, line 79: Use `fs.readSync` instead

package-lock.json
```diff
@@ -0,0 +1,8010 @@
+{
```
Cynosphere, line 1: Vesktop uses pnpm not normal npm.

src/main/index.ts
Cynosphere: Because it doesn't make sense to have a "network request" to grab a local file when you can just read it directly

ading2210:
The changes from this PR got added to #355 so it can be closed now.

PR #590, opened 2024-05-08
chore: update pnpm to latest version

zt64:


Vendicated:
it cannot, no. what's your use case? just use latest version or use corepack

FichteFoll:
It's an issue for packaging on Arch when trying to re-use the pnpm system package: https://aur.archlinux.org/packages/imgbrd-grabber-git

PR #591, opened 2024-05-08
Improve README

Tiagoquix:


PolisanTheEasyNick:
Probably good idea would be to change ScreenSharing screenshot in this PR too, because in #489 we greatly changed UI for ScreenSharing settings on Linux and MacOS systems.

Vendicated:
looks good for the most part but i dislike moving the screenshots down

screenshots should usually be on the top

Curve:
> Current screenshots are too large to be on top, making them a nuisance when trying to scroll down to read about Vesktop's features.
> 
> Resizing them doesn't help because they aren't the same size, so it looks a little out-of-place. That's why I put them at the bottom.
> 
> What do you think about screenshots on top but with something like "Click to expand"?

Put them in a table

README.md
```diff
@@ -1,30 +1,46 @@
...
+### Not yet supported
+
+- Global keybinds.
+- For other things, see the [roadmap](https://github.com/Vencord/Vesktop/issues/324).
+
+## Screenshots
+
+<table>
+<tr>
+<th colspan="3">Linux Screenshare</th>
```
Vendicated, line 25: focuses a bit too much on linux... there should also be one image that just shows normal discord

Issue #588, opened 2024-05-07
Custom launch arguments


SmilerRyan:
### Discord Account

SmilerRyan

### Motivation

Related to issue #23, I'd like to be able to set custom launch arguments for Vesktop without having to make a custom shortcut of Vesktop with my arguments, such as changing the data directory, auto start options, and others. It would also allow the Portable version to be able to be even more portable, by not relying on an extra file on the system (outside the installation directory) for portable builds.

### Solution

Create a launch_arguments.txt file that is built along with all versions of Vesktop.
Portable versions can have the default argument ``--user-data-dir=Data`` as default.
All other versions can have an empty file, and users can manually edit this as they like.

### Alternatives

Another alternative is to not include this file, and in the code just check if this file exists and use the arguments provided.

### Additional context

Here is a mockup of how a portable version would look if arguments were set to put the data folder with the vesktop installation.

I would tag this enhancement instead of bug, however I don't see any option to change this. I think GitHub automatically tagged this as a bug for me.
![image](https://github.com/Vencord/Vesktop/assets/14893385/249479b3-0156-4841-b78d-fc4eb4c9c5b4)

### Request Agreement

- [X] I have searched the existing issues and found no similar issue
- [X] This is not a plugin request
- [X] This is not a Vencord feature request

Vendicated:
your use case seems to just be a portable data dir for portable installs

the logic for portable installs just needs to be fixed to use like `DirectoryOfVesktopExe/data`, it doesnt make much sense outside of portable installs imo

SmilerRyan:
Correct, my use case is to be able to store the data with the application itself, making it use a fixed data directory if it exists.
Not only does this help isolate data more, it also makes running multiple instances and backups easier for me as well.

Vendicated:
since the main motivation has been resolved now, i will be closing this. like i said, there is no good way to do this in a cross platform way

PR #592, opened 2024-05-08
Windows Portable: store settings in portable folder

### LINKED ISSUE #23: Portable Settings
I added a portable build in https://github.com/Vencord/Desktop/commit/6a67f07a9e506a442fb0efdae09dc28cb757a9b9, so you don't have to install it if you don't want to

However, as of now it will still write settings to AppData/Roaming. It would be cool to make it write its settings to the directoy the EXE is in or smth like that

Should be trivial by just using the `PORTABLE_EXECUTABLE_DIR` environment variable, see https://www.electron.build/configuration/nsis#portable

MrGarlic1:
Resolves #23 and #588 

Since the portable build of Vesktop is now a .zip, the previous portable environment variable does not work anymore. Instead, this checks if the platform is windows, and then checks for the presence of `"Uninstall Vesktop.exe"` in the Vesktop install (somewhat sketchy but it works). If the installation is portable, `DATA_DIR` is set to be within the portable install so that settings are persistent.

Also patched the updater which would previously download the non-portable .exe instead of the .zip also due to the portable environment variable issue.


src/main/constants.ts
```diff
@@ -6,10 +6,16 @@
 
 import { app } from "electron";
 import { existsSync, readdirSync, renameSync, rmdirSync } from "fs";
-import { join } from "path";
+import { join, resolve } from "path";
+
+const vesktopDir = resolve(__dirname, "..", "..", "..", "..");
```
Vendicated, line 11: this seems really hacky and platform dependant

maybe try using something like app.getAppPath() or process.execPath

src/updater/main.ts
```diff
@@ -23,17 +24,13 @@ let updateData: UpdateData;
...
     const url = (() => {
         switch (process.platform) {
             case "win32":
                 return assets.find(a => {
-                    if (!a.name.endsWith(".exe")) return false;
-
-                    const isSetup = a.name.includes("Setup");
-                    return portable ? !isSetup : isSetup;
+                    if (PORTABLE && a.name.endsWith("win.zip")) return true;
+                    if (!PORTABLE && a.name.endsWith(".exe")) return true;
```
Vendicated, line 33: this block should properly return a boolean instead of undefined
```suggestion
                    return PORTABLE ? a.name.endsWith("win.zip") : a.name.endsWith(".exe");
```

MrGarlic1: Same logic either way, but that's definitely a clever and precise way to put it! Change made.

src/main/firstLaunch.ts
```diff
@@ -37,6 +37,8 @@ export function createFirstLaunchTour() {
 
     makeLinksOpenExternally(win);
 
+    if (PORTABLE && !existsSync(DATA_DIR)) mkdirSync(DATA_DIR);
```
Vendicated, line 40: i think this call would make more sense directly under where DATA_DIR is defined

also `if (!exists(file)) create(file)` is an antipattern since it could potentially lead to a race condition if the file is deleted/created by a different process between these two checks. it's better to just yolo the action and handle errors. or in the case of mkdir, you can just use the recursive option instead:
```suggestion
    mkdirSync(DATA_DIR, { recursive: true });
```

src/main/constants.ts
```diff
@@ -6,10 +6,16 @@
 
 import { app } from "electron";
 import { existsSync, readdirSync, renameSync, rmdirSync } from "fs";
-import { join } from "path";
+import { dirname, join } from "path";
+
+const vesktopDir = dirname(process.execPath);
```
Vendicated, line 11: this already looks a lot better, but i think when running unpackaged (eg via pnpm start), this will correspond to the electron binary which will lead to it thinking its portable

which technically it is but it would probably be better to treat it as normal? might be as simple as checking if process.execPath is "electron"

Vendicated:
i have made some changes, could you please verify that it still works correctly?

MrGarlic1:
> i have made some changes, could you please verify that it still works correctly?

Just checked, works as expected on Windows 10/11.

Vendicated:
awesome, thank you!

PR #598, opened 2024-05-10
fix: do not re-position on screen that was disconnected

### LINKED ISSUE #597: [Bug] window goes offscreen
### Discord Account

LunarcatOwO

### Operating System

Windows 11

_No response_

if i close vesktop (end it ccompleatly) while its on a second screen then disconnect that screen the next time I start vesktop it will just go off screen and is unable to be retrived

Expect vesktop to snap back on to main screen.

close vesktop (end it ccompleatly) while its on a second screen then disconnect that screen the next time I start vesktop it will just go off screen and is unable to be retrived

### Debug Logs

```
Replace this text with your crash-log. Do not remove the backticks
```
Do not have any

### Request Agreement

MrGarlic1:
Resolves #597. Adds a display ID in the state.json file which will then be used to determine if the previously used display is present. If the display is not present, the window will use default x & y values to position itself instead of attempting to position itself out of bounds.

~~Why Windows doesn't check for this I'm not sure~~

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -12,6 +12,7 @@ import {
     Menu,
     MenuItemConstructorOptions,
     nativeTheme,
+    screen,
     Tray
 } from "electron";
 import { rm } from "fs/promises";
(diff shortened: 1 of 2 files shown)


Vendicated:
thanks, looks good! have you verified that this also works correctly on linux?

MrGarlic1:
Yes, this works on W10/11 and I tested it on Fedora 40 KDE.

Vendicated:
awesome

PR #607, opened 2024-05-16
fix: apply minimize to tray setting correctly on first launch

MrGarlic1:
If the minimize to tray setting was disabled in the first launch window, it would not correctly apply unless it was changed afterwards. Fixed by removing the `!== false` check as it would incorrectly evaluate if the setting was `undefined`. Tested on Windows 10 & Fedora 40 KDE.


Vendicated:
imo this is not the right way to fix that

instead first launch should properly ensure it is always set to a boolean and not undefined

Vendicated:
superseded via https://github.com/Vencord/Vesktop/commit/5fa4ae982410c882ac1bc3fe07318731a5abf1b1

PR #609, opened 2024-05-17
feat: Toggle voice status using fifo pipe

PolisanTheEasyNick:
Note: try to use #326 first.
This PR adds temporarily implementing basic keybinding for toggling mute and deaf statuses.

## How it works?
This PR creates pipe file at `$XDG_RUNTIME_DIR/vesktop-ipc` which listens for strings `VCD_TOGGLE_SELF_MUTE` and `VCD_TOGGLE_SELF_DEAF` and performs actions if this data is transfered to pipe.
So with this if anywhere in the system you execute `echo "VCD_TOGGLE_SELF_MUTE" >> $XDG_RUNTIME_DIR/vesktop-ipc` then it will toggle mute. `echo "VCD_TOGGLE_SELF_DEAF" >> $XDG_RUNTIME_DIR/vesktop-ipc` will toggle self deafness.
## How we can use it?
For example, in KDE systems, we can add custom keybinds which will execute a given command with needed signal:
image
I think in other DEs there must be ways to do similar actions.
If it will be merged, I think we must write these instructions into README. 
This solution is really temporary and in future we must move to some native solutions for keybindings.


src/main/screenShare.ts
```diff
@@ -10,7 +10,7 @@ import { IpcEvents } from "shared/IpcEvents";
 
 import { handle } from "./utils/ipcWrappers";
 
-const isWayland =
+export const isWayland =
```
Vendicated, line 13: would be better and more semantic to move this export to the constants module

src/renderer/index.ts
```diff
@@ -59,3 +59,13 @@ VesktopNative.arrpc.onActivity(async data => {
...
+
+const VoiceActions = findByPropsLazy("toggleSelfMute");
+
+VesktopNative.voice.onToggleSelfMute(() => {
+    VoiceActions.toggleSelfMute();
+});
+
+VesktopNative.voice.onToggleSelfDeaf(() => {
+    VoiceActions.toggleSelfDeaf();
+});
```
Vendicated, line 71:
```suggestion
VesktopNative.voice.onToggleSelfMute(() => VoiceActions.toggleSelfMute());

VesktopNative.voice.onToggleSelfDeaf(() => VoiceActions.toggleSelfDeaf());
```

src/main/mainWindow.ts
```diff
@@ -467,6 +469,7 @@ const runVencordMain = once(() => require(join(VENCORD_FILES_DIR, "vencordDeskto
 export async function createWindows() {
     const startMinimized = process.argv.includes("--start-minimized");
     const splash = createSplashWindow(startMinimized);
+    process.title = "Vesktop";
```
Vendicated, line 472: pretty sure we don't need this, right?
```suggestion
```

src/main/keybinds.ts
```diff
@@ -0,0 +1,41 @@
...
+        const pipe = new net.Socket({ fd });
+        pipe.on("data", data => {
+            const Actions = new Set([IpcEvents.TOGGLE_SELF_DEAF, IpcEvents.TOGGLE_SELF_MUTE]);
+            const action = data.toString().trim();
+            if (Actions.has(action as IpcEvents)) {
+                mainWin.webContents.send(action);
+            }
+        });
+
+        pipe.on("end", () => {
```
Vendicated, line 36:
```suggestion
        pipe.once("end", () => {
```

src/main/keybinds.ts
```diff
@@ -0,0 +1,41 @@
...
+            pipe.destroy();
+            initKeybinds();
```
Vendicated, line 38: do we need to close the file descriptor here?

src/main/keybinds.ts
```diff
@@ -0,0 +1,41 @@
...
+    child_process.spawnSync("mkfifo", [socketFile]);
+    fs.open(socketFile, fs.constants.O_RDONLY | fs.constants.O_NONBLOCK, (err, fd) => {
+        if (err) {
+            console.error("Error opening pipe:", err);
+            return;
+        }
+
+        const pipe = new net.Socket({ fd });
+        pipe.on("data", data => {
+            const Actions = new Set([IpcEvents.TOGGLE_SELF_DEAF, IpcEvents.TOGGLE_SELF_MUTE]);
```
Vendicated, line 29: this should be moved to the top level

src/main/keybinds.ts
```diff
@@ -0,0 +1,41 @@
...
+import net from "net";
+import { IpcEvents } from "shared/IpcEvents";
+
+import { mainWin } from "./mainWindow";
+
+const xdgRuntimeDir = process.env.XDG_RUNTIME_DIR || process.env.TMP || "/tmp";
+const socketFile = path.join(xdgRuntimeDir, "vesktop-ipc");
+
+export function initKeybinds() {
+    child_process.spawnSync("mkfifo", [socketFile]);
```
Vendicated, line 20: this could potentially throw for various reasons, as could fs.open

please wrap it in try catch and add reasonable logging

you can try the error handling by passing bad paths like `/pipe`

src/main/keybinds.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Vesktop, a desktop app aiming to give you a snappier Discord Experience
+ * Copyright (c) 2023 Vendicated and Vencord contributors
+ */
+
+import child_process from "node:child_process";
+import fs from "node:fs";
+import path from "node:path";
+
+import net from "net";
+import { IpcEvents } from "shared/IpcEvents";
```
Vendicated, line 12: always prefer picking only the imports you need instead of importing the entire module

```suggestion
import { spawnSync } from "node:child_process";
import { open } from "node:fs";
import { join } from "node:path";

import { Socket } from "net";
import { IpcEvents } from "shared/IpcEvents";
```

src/main/keybinds.ts
PolisanTheEasyNick: i believe that yes because after user ended sending info to pipe (when echo ends work) then it just stops receiving further messages, so i need to recreate pipe
i will search for more clean ways

Vendicated:
this doesn't really work still. the first command works, but then runs into
```
Failed to mkfifo /run/user/1000/vesktop-ipc while initializing vesktop-ipc: Error: Command failed: mkfifo /run/user/1000/vesktop-ipc
mkfifo: cannot create fifo '/run/user/1000/vesktop-ipc': File exists
```

PolisanTheEasyNick:
Didn't have this issue, but now before creating a pipe file it checks whether it exists, if so removes the file and recreates. Also, it must now remove the file if Vesktop is closed.

hollowstrawberry:
Works great for me.

minchingtonak:
Hi there, thanks for creating this feature. I'm using it now and it works nicely. I did have to patch the implementation to allow non-wayland display servers though. I'm curious, is there a reason the keybinds are only enabled on wayland? It seems to be working fine and I'm running X11.

JoshIce:
Hey just wanted to say this is working great! Hoping we can get this or some other solution merged into main soon!

CookieUniverse:
I don't really know if its my problem but when run echo "VCD_TOGGLE_SELF_MUTE" >> $XDG_RUNTIME_DIR/vesktop-ipc it just hangs the console and nothing happens.
Using the latest Nobara Linux with wayland and nvidia drivers if that matters.
image

PolisanTheEasyNick:
> I don't really know if its my problem but when run echo "VCD_TOGGLE_SELF_MUTE" >> $XDG_RUNTIME_DIR/vesktop-ipc it just hangs the console and nothing happens. Using the latest Nobara Linux with wayland and nvidia drivers if that matters.

Please show output of `echo $XDG_RUNTIME_DIR` and `ls -la $XDG_RUNTIME_DIR`

CookieUniverse:
> > I don't really know if its my problem but when run echo "VCD_TOGGLE_SELF_MUTE" >> $XDG_RUNTIME_DIR/vesktop-ipc it just hangs the console and nothing happens. Using the latest Nobara Linux with wayland and nvidia drivers if that matters.
> 
> Please show output of `echo $XDG_RUNTIME_DIR` and `ls -la $XDG_RUNTIME_DIR`

image

CookieUniverse:
turns out my problem wasn't related to this. this works!

ilobilo:
Hi is this getting merged any time soon?

PolisanTheEasyNick:
> Hi is this getting merged any time soon?

I don't believe so, this PR is desired for other users who needs that feature.

ilobilo:
do I need to build it myself to use it?

minchingtonak:
> do I need to build it myself to use it?

Yes. Use the instructions [here](https://github.com/Vencord/Vesktop?tab=readme-ov-file#building-from-source), make sure to check out this branch right after cloning (`git switch keybinds`).

src/main/mainWindow.ts
```diff
@@ -512,4 +514,5 @@ export async function createWindows() {
     });
 
     initArRPC();
+    if (isWayland) initKeybinds();
```
minchingtonak, line 517: Why are keybinds only enabled for wayland users? I'm an X11 user and keybinds work just fine from my local testing when I remove the conditional.

HateJoysticks:
I think I'm doing something wrong here, Can someone explain how to do it?

ilobilo:
could this branch be updated with main?

vinivosh:
Just saying that this is working **perfectly** for my use case! I'm impressed, hope this gets merged some time.

mgabor3141:
Hi, I've resolved the conflicts and updated the branch with the changes from main.
See: https://github.com/mgabor3141/Vesktop/pull/1 or https://github.com/PolisanTheEasyNick/Vesktop/pull/2

I've also added `MUTE` and `UNMUTE` commands in addition to the `TOGGLE_MUTE` command.
See: [mgabor3141/Vesktop@`328710d` (#2)](https://github.com/mgabor3141/Vesktop/pull/2/commits/328710d112d1aba6dbb3b7b9431938a9c1be244d)

Vendicated:
Superseded by https://github.com/Vencord/Vesktop/pull/1238 which uses a unix socket server instead

PR #621, opened 2024-05-24
linux audioshare: add granular selection, more options, better ui

### LINKED ISSUE #218: add support for blacklisting apps for audio screenshare
usually, when i'm in a discord call, it's with my partner, and both of us deafen each other since we speak over mumble and use discord's screensharing feature to watch youtube videos or play games. 

in order to prevent having to close the screenshare and re-open a new one whenever either of us decide to do something else, we usually use the "entire system" option to capture all audio all of the time, and then i have to open helvum (pipewire patchbay) to disconnect mumble from vesktop, so that my partner can't hear himself.

i noticed that vesktop automatically excludes itself from its sound being streamed over screenshare, presumably to prevent echo from the voice chat. if vesktop has this ability, maybe there can be an option to automatically exclude *any* app the user wants (in my case, mumble)? it could maybe be a text box where you can enter the names of binaries to exclude, or even maybe a dropdown list that lists all of the current running apps and lets you pick which ones to opt out of sharing.

### LINKED ISSUE #619: [Linux][X11] Audio Sinks as Stream Input
### Discord Account

apostle_of_floof

### Motivation

When selecting an audio source for a screen share, you can only select "Entire System" or individual apps.
Entire Audio obviously isn't exactly the greatest choice when using sinks, as it even connects the sinks discord is piped into (meaning viewers will hear themselves in the stream).
Entire System

The second option of selecting individual apps isn't working as great either. Some apps/games just don't appear in the list (E.g. Rabbit and Steel), and some other games take ALL nodes related to the game, including input/microphone nodes.
Helldivers

### Solution

Enable selecting audio sinks and/or indidivual devices as audio source for streams, similar to how it works in discords voice settings.

### Alternatives

I tried manually setting the connections in qpwgraph, but starting a new stream/changing the audio source will reset all connections, forcing me to manually set/remove each connection again. Exclusive Mode fixes this issue, but forces you to connect new nodes manually to devices and such.

### Additional context

_No response_

### Request Agreement

- [X] This is not a plugin request
- [X] This is not a Vencord feature request

Curve:
- [x] Bump Venmic
- [x] Granular Selection
- [x] Allow to select multiple audio sources
- [x] Additional Exclusion Options
- [x] Redesign Screen-Share Picker
- [x] New "Ignore Virtual" Setting
- [x] Add missing venmic settings
- [x] Make AudioShare settings persistent 

---

image
image
image
image

---

> [!NOTE]
> The settings in the screenshot do not represent the defaults :)

Fixes #619
Fixes #218

### DIFF
--- a/src/main/venmic.ts
+++ b/src/main/venmic.ts
@@ -4,15 +4,13 @@
  * Copyright (c) 2023 Vendicated and Vencord contributors
  */
 
-import type { Node, PatchBay as PatchBayType } from "@vencord/venmic";
+import type { LinkData, Node, PatchBay as PatchBayType } from "@vencord/venmic";
 import { app, ipcMain } from "electron";
 import { join } from "path";
 import { IpcEvents } from "shared/IpcEvents";
 import { STATIC_DIR } from "shared/paths";
 
 import { Settings } from "./settings";
-
-type LinkData = Parameters<PatchBayType["link"]>[0];
 
 let PatchBay: typeof PatchBayType | undefined;
 let patchBayInstance: PatchBayType | undefined;
@@ -85,17 +83,17 @@ ipcMain.handle(IpcEvents.VIRT_MIC_START, (_, include: Node[]) => {
     const { ignoreDevices, ignoreInputMedia, ignoreVirtual, workaround } = Settings.store.audio ?? {};
 
     const data: LinkData = {
-        include: include,
+        include,
         exclude: [{ "application.process.id": pid }],
         ignore_devices: ignoreDevices
     };
 
     if (ignoreInputMedia ?? true) {
-        data.exclude!.push({ "media.class": "Stream/Input/Audio" });
+        data.exclude.push({ "media.class": "Stream/Input/Audio" });
     }
 
     if (ignoreVirtual) {
-        data.exclude!.push({ "node.virtual": "true" });
+        data.exclude.push({ "node.virtual": "true" });
     }
 
     if (workaround) {
@@ -112,6 +110,7 @@ ipcMain.handle(IpcEvents.VIRT_MIC_START_SYSTEM, (_, exclude: Node[]) => {
         Settings.store.audio ?? {};
 
     const data: LinkData = {
+        include: [],
         exclude: [{ "application.process.id": pid }, ...exclude],
         only_speakers: onlySpeakers,
         ignore_devices: ignoreDevices,
--- a/src/main/venmic.ts
+++ b/src/main/venmic.ts
@@ -4,7 +4,7 @@
  * Copyright (c) 2023 Vendicated and Vencord contributors
  */
 
-import type { PatchBay as PatchBayType } from "@vencord/venmic";
+import type { Node, PatchBay as PatchBayType } from "@vencord/venmic";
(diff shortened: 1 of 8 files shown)


Curve:
> i think it would also be useful to update the Pulseaudio server warning to add a recommendation to switch to pipewire via https://gist.github.com/the-spyke/2de98b22ff4f978ebf0650c90e82027e#install

Will do!

src/main/venmic.ts
```diff
@@ -69,47 +71,59 @@ function getRendererAudioServicePid() {
...
+        ignore_devices: settings.audioIgnoreDevices
     };
 
-    if (workaround) {
-        data.workaround = [
-            { key: "application.process.id", value: pid },
-            { key: "media.name", value: "RecordStream" }
-        ];
+    if (settings.audioIgnoreInputMedia ?? true) {
+        data.exclude?.push({ "media.class": "Stream/Input/Audio" });
```
Vendicated, line 92:
```suggestion
        data.exclude.push({ "media.class": "Stream/Input/Audio" });
```

src/main/venmic.ts
```diff
@@ -69,47 +71,59 @@ function getRendererAudioServicePid() {
...
+    }
+
+    if (settings.audioIgnoreVirtual) {
+        data.exclude?.push({ "node.virtual": "true" });
```
Vendicated, line 96:
```suggestion
        data.exclude.push({ "node.virtual": "true" });
```

src/main/venmic.ts
```diff
@@ -69,47 +71,59 @@ function getRendererAudioServicePid() {
...
+        data.workaround = [{ "application.process.id": pid, "media.name": "RecordStream" }];
     }
 
     return obtainVenmic()?.link(data);
 });
 
-ipcMain.handle(IpcEvents.VIRT_MIC_START_SYSTEM, (_, workaround?: boolean, onlyDefaultSpeakers?: boolean) => {
+ipcMain.handle(IpcEvents.VIRT_MIC_START_SYSTEM, (_, exclude: Node[]) => {
     const pid = getRendererAudioServicePid();
+    const settings = Settings.store;
```
Vendicated, line 108: why not use destructuring?
```suggestion
    const { audioOnlySpeakers, audioIgnoreDevices, ... } = Settings.store;
```

src/shared/settings.d.ts
```diff
@@ -30,6 +30,16 @@ export interface Settings {
...
+
+    audioWorkaround?: boolean;
+    audioGranularSelect?: boolean;
+
+    audioIgnoreVirtual?: boolean;
+    audioIgnoreDevices?: boolean;
+    audioIgnoreInputMedia?: boolean;
+
+    audioOnlySpeakers?: boolean;
+    audioOnlyDefaultSpeakers?: boolean;
```
Vendicated, line 42:
```suggestion
    venmic: {
    	workaround?: boolean;
    	granularSelect?: boolean;
    	ignoreVirtual?: boolean;
    	ignoreDevices?: boolean;
    	ignoreInputMedia?: boolean;
    	onlySpeakers?: boolean;
    	onlyDefaultSpeakers?: boolean;
    }
```

PR #635, opened 2024-05-31
Update build choosing instructions for Macs

Unichy:
Picking the Intel build to run on an Apple Silicon Mac is not the best advice. In lieu of a Universal build, this updates the instructions directing the user to check their architecture under About This Mac from the Apple menu.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ If you don't know the difference, pick the Installer.
 
 ### Mac
 
-If you don't know the difference, pick the Intel build.
+If you don't know the difference, check About This Mac from the Apple menu. If the first item is labeled "Chip", choose the Apple Silicon build. If it is labeled "Processor", choose the Intel build. 
 
 - [Intel build (amd64)](https://vencord.dev/download/vesktop/amd64/dmg)
 - [Apple Silicon (arm64)](https://vencord.dev/download/vesktop/arm64/dmg)


README.md
```diff
@@ -26,7 +26,7 @@ If you don't know the difference, pick the Installer.
 
 ### Mac
 
-If you don't know the difference, pick the Intel build.
+If you don't know the difference, check About This Mac from the Apple menu. If the first item is labeled "Chip", choose the Apple Silicon build. If it is labeled "Processor", choose the Intel build.
```
Vendicated, line 29: why not just make them run a terminal command like `uname -m` or `arch`? way simpler and less confusing

Unichy: It could work but might be intimidating for beginners.

Vendicated: it really shouldn't be. just need to open terminal and copy paste command

Unichy: I’m not sure if using the terminal is simpler. You can get to the about this Mac menu in one click. No need to copy and paste. 

Either way, the user has to decipher the info and choose the correct build for themselves.

TheScreechingBagel: the overlap for users who don't know whether they have an apple silicon mac and users who will be intimidated by a terminal is probably high haha

splatterxl:
> Picking the Intel build to run on an Apple Silicon Mac is not the best advice. 

How? If a user is not running Apple Silicon the [Rosetta software](https://en.wikipedia.org/wiki/Rosetta_(software)) will run Intel builds anyway. However on Intel macs arm64 will not work at all.

Vendicated:
Vesktop 1.5.3 is now a universal build so this is no longer needed

Issue #600, opened 2024-05-12
[Bug] No application icon present in LXDE


ading2210:
### Discord Account

vk6

### Operating System

Debian 12

### Linux Only ~ Desktop Environment

LXDE (uses Xorg)

### What happens when the bug or crash occurs?

When running Vesktop inside the LXDE desktop environment, there is no app icon.

![image](https://github.com/Vencord/Vesktop/assets/71154407/43fe9e80-4527-410a-ab56-95ccf92c808b)
This may be related to #575 but it seems like that one is related to the AppImage builds only. 

### What is the expected behaviour?

I expect the app icon to be the same as what is shown in the tray icon. I believe this is also what the behavior is on other desktop environments.

![image](https://github.com/Vencord/Vesktop/assets/71154407/dba13031-c5c4-45c6-9dca-81aef4e38c32)

### How do you recreate this bug or crash?

1. Launch LXDE
2. Install Vesktop with the .deb file
3. Launch Vesktop
4. Observe the task bar

### Debug Logs

```
[arRPC > ipc] listening at /run/user/1000/discord-ipc-0
[arRPC > websocket] listening on 6463
[arRPC > process] started
[Vencord] Installed React Developer Tools
(node:22912) ExtensionLoadWarning: Warnings loading extension at /home/allen/.config/vesktop/ExtensionCache/fmkadmapgofadopljbjfkapdkoienihi:
  Manifest version 2 is deprecated, and support will be removed in 2024. See https://developer.chrome.com/docs/extensions/develop/migrate/mv2-deprecation-timeline for details.

(Use `vesktop --trace-warnings ...` to show where the warning was created)

(vesktop:22912): libnotify-WARNING **: 01:02:50.380: Failed to connect to proxy

(vesktop:22912): libnotify-WARNING **: 01:02:50.380: Failed to connect to proxy

(vesktop:22912): libnotify-WARNING **: 01:02:50.381: Failed to connect to proxy
[22912:0512/010250.381894:ERROR:libnotify_notification.cc(50)] notify_notification_show: domain=684 code=2 message="GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.Notifications was not provided by any .service files"
```

### Request Agreement

- [X] I have searched the existing issues and found no similar issue
- [X] I am using the latest Vesktop and Vencord versions
- [X] This issue occurs on an official release (not just the AUR or Nix packages)

thomas-phillips:
[...]

PR #643, opened 2024-06-06
Changed icon size from 1080x1080 to 512x512

thomas-phillips:
The icon file size was changed to 512x512 as recommended in the electron builder's documentation regarding icon size.

This addresses the #600 issue regarding broken icon size in the Debian 12 LXDE desktop environment.

The icon size electron builder recommends for all platforms is 212x212 or 512x512.

20240606_15h46m15s_grim
20240606_15h46m26s_grim


Vendicated:
did you mean to replace the icon in build/ ?

thomas-phillips:
> did you mean to replace the icon in build/ instead?

Changing the size of the icon in build/ unfortunately doesn't fix the issue. Could it be as icons are  sourced from static/ [here](https://github.com/Vencord/Vesktop/blob/abd1e8c6daed8bd5de41c049813bc08113384ba1/src/shared/paths.ts#L12) that the electron-builder is ignoring icons in build/ for linux builds

Vendicated:
the icon in static is only used for tray, nothing else

the icon used in build is the one used as application icon

thomas-phillips:
That makes sense, would it be okay to reduce the tray icon size?

Vendicated:
should be resolved via https://github.com/Vencord/Vesktop/commit/6e2da1d29472c452548f47500e635af7ea7007bf

Issue #424, opened 2024-03-06
Add help text to Notification Badge option


Tiagoquix:
**Is your feature request related to a problem? Please describe.**
The Notification Badge option only works when `libunity` is installed, but this might not be exactly clear to non-Ubuntu users.

**Describe the solution you'd like**
In the option's description, add a help text such as:
![image](https://github.com/Vencord/Vesktop/assets/30274161/38198b24-6931-4978-bb5d-3969ecbfbfa6)
`Note: if this is not working for you, be sure to install the "libunity" package for your distro.`

**Describe alternatives you've considered**
Declare `libunity` as a dependency.

**Additional context**
I'm on Fedora 39 KDE and had to install it separately.

Vendicated:
it should just be added as a dependency

PR #686, opened 2024-06-21
feat(linux): Emit D-Bus signal directly for dock notification badges

### LINKED ISSUE #298: Unread Notification Badge [Linux]
Cross-platform duplicate of [this Win10 feature request](https://github.com/Vencord/Vesktop/issues/20).

Feature looks like this:
notif_badge.

Note: Vanilla discord has a missing dependency in libunity which is responsible for notification badges, as referenced [in this thread](<https://github.com/flathub/com.discordapp.Discord/issues/228>).
As shown above, the feature works on vanilla discord after installing libunity (or on distros that come with it), but not on Vencord.

Covkie:
Closes #298, #424, #189

remove (#) title prefix when `Notification Badge` option is toggled

explode scuffed optional libunity dependency (kabloomed)

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -385,6 +385,8 @@ function initSpellCheck(win: BrowserWindow) {
 }
 
 function initTitlePatches(win: BrowserWindow) {
+    // let listener: boolean;
+
     const updateTitle = (e: { preventDefault: Function }, title: string) => {
         if (Settings.store.staticTitle) {
             win.setTitle("Vesktop");
@@ -395,6 +397,21 @@ function initTitlePatches(win: BrowserWindow) {
         }
     };
     win.on("page-title-updated", updateTitle);
+    // const manageListeners = (enabled: boolean | undefined) => {
+    //     const active = Settings.store.staticTitle || Settings.store.appBadge;
+
+    //     if ((enabled || active) && !listener) {
+    //         win.on("page-title-updated", updateTitle);
+    //         listener = true;
+    //     } else if (!enabled && !active && listener) {
+    //         win.off("page-title-updated", updateTitle);
+    //         listener = false;
+    //     }
+    // };
+
+    // manageListeners(true);
+    // addSettingsListener("staticTitle", manageListeners);
+    // addSettingsListener("appBadge", manageListeners);
 }
 
 function createMainWindow() {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,7 +6,7 @@ settings:
 
 patchedDependencies:
   arrpc@3.5.0:
-    hash: sczviclcirm3d4e2soabfniygq
+    hash: 4313fe844324a52ef0453078a86f5d1ee2d4f406331201211020d9fa243323c4
     path: patches/arrpc@3.5.0.patch
 
 importers:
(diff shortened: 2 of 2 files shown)


src/main/appBadge.ts
```diff
@@ -25,7 +26,18 @@ export function setBadgeCount(count: number) {
     switch (process.platform) {
         case "linux":
             if (count === -1) count = 0;
-            app.setBadgeCount(count);
+            // app.setBadgeCount(count);
+            
+            function emitDBusBadge(count: number, visible: boolean) {
+                const badgeCountCommand = `gdbus emit --session --object-path / --signal com.canonical.Unity.LauncherEntry.Update "application://vesktop.desktop" "{'count': <int64 ${count}>, 'count-visible': <${visible}>}"`;
```
Vendicated, line 32: since count comes from renderer via ipc, it is an untrusted value. embedding it into an `exec` command is a veeery bad idea. what if the browser passes something like `"$(rm -rf /*)"`?

you absolutely need to validate the value to avoid sandbox escapes, e.g. via `typeof count === "number"`

Vendicated: also raw exec with string is not very robust, prefer calls that take an arg array, e.g. `execFile("gdbus", ["emit", ...])`. that's also more performant since it won't spawn an extra shell

ToxicMushroom:
> Do I provide an unread badge at all and if I do do i make it display a "-" or "0" (also should I strip out the `•` like with the `(#)`

I think it makes most sense to not display an unread pings badge if the count is 0.
So it makes it more noticeable when you do have one.

Idm either way for the •

src/main/mainWindow.ts
```diff
@@ -445,6 +445,10 @@ function createMainWindow() {
     });
 
     if (Settings.store.staticTitle) win.on("page-title-updated", e => e.preventDefault());
+    else if (Settings.store.appBadge) mainWin.webContents.on('page-title-updated', (_, title) => {
+        let cleanedTitle = title.replace(/^\(\d+\)\s*|•\s/, '');
+        mainWin.setTitle(cleanedTitle);    
+    });
```
Vendicated, line 451:
```suggestion
    else if (Settings.store.appBadge)
    	mainWin.webContents.on("page-title-updated", (_, title) => {
	        title = title.replace(/^\(\d+\)\s*|•\s/, "");
	        mainWin.setTitle(cleanedTitle);    
	    });
```

src/renderer/appBadge.ts
```diff
@@ -11,17 +11,19 @@ import { Settings } from "./settings";
...
 
     try {
         const mentionCount = GuildReadStateStore.getTotalMentionCount();
         const pendingRequests = RelationshipStore.getPendingCount();
+        const messageRequests = MessageRequestStore.getMessageRequestsCount();
         const hasUnread = GuildReadStateStore.hasAnyUnread();
         const disableUnreadBadge = NotificationSettingsStore.getDisableUnreadBadge();
 
-        let totalCount = mentionCount + pendingRequests;
+        let totalCount = mentionCount + pendingRequests + messageRequests;
```
Vendicated, line 26: why? this code is ported exactly from Discord. they don't include this either. we should keep behaviour consistent.

image

Vendicated (changes requested):

src/main/appBadge.ts
```diff
@@ -24,8 +25,24 @@ let lastIndex: null | number = -1;
 export function setBadgeCount(count: number) {
     switch (process.platform) {
         case "linux":
-            if (count === -1) count = 0;
```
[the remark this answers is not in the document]
Vendicated: but you just moved it down? and it breaks the `count !== 0` check
image

flying-sheep:
Why gdbus? It doesn’t necessarily exist on platforms supporting this DBus API (e.g. KDE Plasma)

Can you switch to a DBus node.js library?

flying-sheep:
`dbus-send` is probably the better call: on Arch Linux, Ubuntu, Debian, and Fedora it’s provided by the `dbus` package, which is installed on every system with a DE, and depended on by many many things.

There are two DBus implementations: dbus-daemon (included with `dbus`) and dbus-broker. There are a few systems that *only* use dbus-broker, in which case `dbus-send` is not necessarily present. That currently includes only immutable distros like Fedora Silverblue. Debian has a split-off `dbus-bin` package that contains `dbus-send` so maybe going forward immutable distros also come with it.

So all in all I’d say if you go with one tool, `dbus-send` is the safer bet, multiple tools would be better, and a node.js library would be the perfect, always-works alternative.

src/main/appBadge.ts
```diff
@@ -24,8 +25,21 @@ let lastIndex: null | number = -1;
...
+            }
+            
+            execFile ("gdbus", [
+                "emit",
+                "--session",
+                "--object-path",
+                "/",
+                "--signal",
+                "com.canonical.Unity.LauncherEntry.Update",
+                "application://vesktop.desktop",
```
JakobDev, line 39: Please note that you need to use `dev.vencord.Vesktop.deskto`p if running as Flatpak.

CarJem:
Curious where this PR stands as I'd like to have this fix already so I can ditch the Official Discord Client on GNOME Wayland.

Michael-B8:
Hey, I've also built with this since my notification badge doesn't work without this PR. I've also ran into @RealSweetPanda's problem (and I'm on Plasma), in which one of my dbus-broker processes reaches its max file descriptor limit (1024) then it crashes and so does everything on my pc. It seems every single call to `dbus.sessionBus()` opens a connection that's never closed.

I think I managed to fix it on my machine (didn't test for too long, but seems to work and not leave too many connections hanging) by making `sessionBus` a singleton. I don't know much about js though so I don't know if that's a good solution lol.

Michael-B8:
> @Michael-B8 can you lmk if that fixed it.

Yep, doesn't seem to create any extra connections anymore. Thanks :)

If you're curious how I went about checking, I got the pid of the main vesktop process `pgrep vestkop -a` (or electron), then checked the pid of every dbus-broker `pgrep dbus-broker` for vesktop opened files `sudo lsof -p <pid_dbus> | grep <pid_vesktop>`. And when I found the one that has opened files I message my discord account to trigger a notification badge update and check if the number of vesktop opened files stayed the same. Before, it added a connection for every badge update and eventually it reached the limit and crashed.

package.json
```diff
@@ -28,7 +28,8 @@
         "electron-updater": "^6.3.4"
     },
     "optionalDependencies": {
-        "@vencord/venmic": "^6.1.0"
+        "@vencord/venmic": "^6.1.0",
+        "@homebridge/dbus-native": "0.6.0"
```
Vendicated, line 32: adding a new library that adds more than 200kb of javascript just for one single dbus call is pretty bad

Adamcake:
This is the main thing preventing me from switching from flatpak's com.discordapp.Discord to pretty much any other solution, so it would be really nice to have this done and not have to think about libunity ever again. What's holding it up at the moment?

@Covkie 
> Yeah electron fixed (removed) their desktop env checking for the libunity dep. I still think this PR is relevant tho as libunity is old and many distros don't package it anymore. It should be dropped in favour of dbus communication.

This is interesting, and doesn't match my experience, but would definitely be convenient if true. To get notification badges working in com.discordapp.Discord we had to not only figure out how to package libunity but also do some separate thing to convince electron that libunity is actually available.

Vendicated:
superseded by #1180

PR #721, opened 2024-07-04
meta: Use Network category instead of AudioVideo

lionirdeadman:
Messaging applications should use the Network category, not AudioVideo.

See: https://specifications.freedesktop.org/menu-spec/latest/apa.html

---

I noticed this when I saw Vesktop on Flathub in the AudioVideo category where it looks quite out of place.


Vendicated:
honestly no idea why it was ever AudioVideo to begin with, that doesn't make sense to me either. perhaps whoever wrote it considered its voice chat capabilities to fall under that?

thanks for the fix!

PR #737, opened 2024-07-08
Add vesktop icons for linux packagers

### LINKED ISSUE #691: Provide build/icon.icns as a zip for Linux packagers
### Content

The `.icns` format on Linux is not well supported and extracting the pngs from said format is impossible to do fully. 

libicns (the library which provides `icns2png`) has not had a stable release in **12** years and only manages to extract two large icons from the icon.icns. Latest git is a bit better and only fails for ic04 and ic05 but again, doesnt work fully. 

I'm making this as `vesktop`/`-git` on the aur uses `build/icon.png` which is notably *not* the Vesktop icon but Vencord's,.,..

Covkie:
closes #691

Extracted from latest deb


Vendicated:
don't add it as a tar, what in the world

PR #771, opened 2024-07-19
migrate vencord loader to asar

Vendicated:


thororen1234:
Tested it all works good!

BrycensRanch:
What are the benefits of this PR? What does it address?

lewisakura:
> What are the benefits of this PR? What does it address?

confused on why you're asking, but the main vencord mod is moving to asar so this is a necessity for vesktop's continued functionality

lewisakura:
Was unable to confirm if the migration step works (due to my own fault, I'll see if I can fix it later), but the updater seems to pull the correct files from remote and Vesktop seems to be able to load them just fine.

Issue #365, opened 2024-01-25
visibility as desktop instead of web


moonfight1:
**Is your feature request related to a problem? Please describe.**
When using the platform indicators plugin, i appear as web not desktop

**Describe the solution you'd like**
To make the vesktop client be visible as desktop not web server-side 

**Additional context**
![image](https://github.com/Vencord/Vesktop/assets/156691675/b7811057-ed9a-4caa-ab6f-93c5d258a91e)

Vendicated:
no

riverpaper:
can we revisit this? why not make it optional as a plugin?

nin0-dev:
vencord won’t add a potentially bannable plugin, this won’t be “revisited”

PR #784, opened 2024-07-28
feat: Report Vesktop as a desktop client

Covkie:
Adds a patch to show Vesktop as a desktop client for users of Platformindicators. Vesktop aims to be a Discord desktop replacement with all the usual desktop features. Unlike the web(which Vesktop currently reports as) version that can run on different platforms, Vesktop is only for desktops so its useful to report this to platformIndicators :3


D3SOX:
https://github.com/Vencord/Vesktop/issues/365#issuecomment-1910420637

PR #805, opened 2024-08-08
Add Electron 31/Chromium 126 workaround flag

C-512L:
See:https://issues.chromium.org/issues/331796411

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -42,6 +42,12 @@ function init() {
         app.commandLine.appendSwitch("disable-smooth-scrolling");
     }
 
+    // Disable multiplane format for software video decoding on Linux due to bug on Electron 31/Chromium 125
+    // https://issues.chromium.org/issues/331796411
+    if (process.platform == "linux") {
+      disabledFeatures.push("UseMultiPlaneFormatForSoftwareVideo")
+    }
+
     // disable renderer backgrounding to prevent the app from unloading when in the background
     // https://github.com/electron/electron/issues/2822
     // https://github.com/GoogleChrome/chrome-launcher/blob/5a27dd574d47a75fec0fb50f7b774ebf8a9791ba/docs/chrome-flags-for-tools.md#task-throttling


Vendicated:
why does it matter if it logs that or not? this isn't relevant to 99.9% of users. you can just pass it yourself

C-512L:
> why does it matter if it logs that or not? this isn't relevant to 99.9% of users. you can just pass it yourself

First of all, please refrain from using that kind of attitude, is not helpful at all and inflammatory which can drive people away from the project (which is something I don't want to, as I consider this a good project). 

If you want me to explain why this is an issue, please do so in a proper manner, otherwise this can not get to be a fruitful discussion.

Now onto the issue, there are several reasons for fixing Vesktop from doing repeated spam errors:
- It makes vesktop devs/contributors' work harder to browse logs, by filling them with unhelpfull information. Also the "average" linux user knows about checking logs, so this affects him too.
-  It is a real upstream issue on Chromium, which affects this project as a downstream consuner and ignoring the issue doesn't make it disappear.
- On Linux systems using sytemd every message is sent to the sytemd-journal (which also gets the spam) and stored on the persistent storage, making it an unnecesary use of disk space and indirectly reducing its usable life (something an average non-tech user might care)

So please consider taking this pull request as what it is, me trying to improve Vesktop the way I can.

Vendicated:
> First of all, please refrain from using that kind of attitude, is not helpful at all and inflammatory

wha? i was just stating my reason for closing this. it wasn't meant in a negative way, sorry if you interpreted it that way

in any case, I still don't believe any of your points have much relevancy. these logs only show once you watch a stream, so hey have little impact on debugging. i honestly didn't even know this was a thing and I debug Vesktop a lot. the log volume is also gonna be fairly low compared to everything else going on in your system so i doubt it has any relevant performance impact

i have no idea what the feature that flag is disabling does, but I don't think outright disabling a feature for everyone just to avoid log spam that affects 0.01% of users is a good idea. google is gonna have had a reason to add the feature

C-512L:
> wha? i was just stating my reason for closing this. it wasn't meant  
> in a negative way, sorry if you interpreted it that way

Thanks for clarifying it, I suppose I wasn't ill-intended.

> in any case, I still don't believe any of your points have much  
> relevancy. these logs only show once you watch a stream, so hey have  
> little impact on debugging. i honestly didn't even know this was a  
> thing and I debug Vesktop a lot. 

I know at least one friend also affected by this.
The issue affects at least Linux on Wayland with an AMD GPU (only I could test with Intel and AMD).

> the log volume is also gonna be fairly low compared to everything else
> going on in your system so i doubt it has any relevant performance impact

not really. The log gets triggered every time a frame from any video playback (screenshare, youtube embeds, video attachments) is rendered, so a 30fps video triggers it 30 times in a second.

> i have no idea what the feature that flag is disabling does

Disables support for multi-plane image formats which it is optional feature for video decoders. It shouldn't affect the average user.

> don't think outright disabling a feature for everyone just to avoid  
> log spam that affects 0.01% of users is a good idea. 

This patch only disables it for Linux users (both on X11 and Wayland), but after reviewing it a little further I determined that using `--disable-gpu-memory-buffer-video-frames` instead would be a better option. That also should not affect X11 users as chromium has disabled GPU memory buffers for X11 by default for years due to multiple bugs (on Wayland and all other OSes GPU Memory Buffers are used by default). But I would like to know in advance if such a PR would be approved.

> google is gonna have had a reason to add the feature

Google has broken Chromium Linux builds a few times before, specially anything involving hardware acceleration (ie: hardware video decoding has been broken for years and to this day "almost works"). Most of the work done for the Linux build is for ChromeOS, which is a very different environment than the average Linux distribution and also runs on a limited set of hardware.

Issue #769, opened 2024-07-19
[Feature Request] Implement discord:// scheme / protocol


Write:
### Motivation

I'd like to use this app as a discord replacement.
This has been requested here https://github.com/Vencord/Vesktop/issues/200 but closed without any response / reason.

### Solution

Add support for discord:// url-scheme 

### Alternatives

Proxying discord:// to vesktop:// but this doesn't exist either.

### Request Agreement

- [X] I have searched the existing issues and found no similar issue
- [X] This is not a plugin request
- [X] This is not a Vencord feature request

mmaietta:
Quick chime-in since I noticed this project leverages electron-builder. Adding a protocol scheme is very easy in the builder config.
```
  protocols: {
    name: "Vesktop Example",
    schemes: [
      "vesktop"
    ]
  },
  linux: {
    desktop: {
      MimeType: "x-scheme-handler/vesktop"
    },
  },
```

Vendicated:
> This has been requested here https://github.com/Vencord/Vesktop/issues/200 but closed without any response / reason.

I closed it because i decided it wasn't useful enough to warrant the time spent. I haven't seen the protocol used much, not even by Discord. What aspects of it do you consider useful? As in, which routes?

Write:
> > This has been requested here #200 but closed without any response / reason.
> 
> I closed it because i decided it wasn't useful enough to warrant the time spent. I haven't seen the protocol used much, not even by Discord. What aspects of it do you consider useful? As in, which routes?

First example coming to my mind : Every time a server is joined via a web link, it should open the app.

Vendicated:
that already works! :P it's done via RPC, not the protocol

Write:
> that already works! :P it's done via RPC, not the protocol

This doesn't work for me ? It tries to open a discord:// url scheme when I click on "launch app" and it just says no app exist for such scheme, or something along those lines. At least on macOS + Safari.

Guessing RPC would require discord to be already running ?

Vendicated:
Oh i see, interesting! Indeed, RPC only works if you have Vesktop open. I wasn't aware they use the protocol when RPC is unreachable, but you are right

Write:
In the meantime,

I used [Platypus](https://sveinbjorn.org/platypus) to register discord://  scheme and open vesktop app.
It doesn't passtrough args to vesktop but merely just opens the app

Example of config for Platypus :
[...]

PR #813, opened 2024-08-13
Support discord:// uri scheme

Covkie:
closes: #769

Could someone on mac confirm this works as well? (I followed the electron docs so im like pretty sure it works but :woozy_face:)


src/main/mainWindow.ts
```diff
@@ -455,12 +455,20 @@ function createMainWindow() {
...
-        Settings.store.discordBranch === "canary" || Settings.store.discordBranch === "ptb"
-            ? `${Settings.store.discordBranch}.`
-            : "";
+    const branch = Settings.store.discordBranch;
+    const subdomain = branch === "canary" || branch === "ptb" ? `${branch}.` : "";
+    const uri = process.argv.find(arg => arg.startsWith("discord://"));
 
-    win.loadURL(`https://${subdomain}discord.com/app`);
+    const loadUrl = (url: string | undefined) => {
+        win.loadURL(`https://${subdomain}discord.com${url?.substring("discord://".length) || "/app"}`);
```
Sqaaakoi, line 463: Do not assume the host is an empty string, often times Discord URIs are `discord://-/`

Sqaaakoi (changes requested):

ryanamay:
Confirmed Working on macOS Sonoma 14.6 via `open discord://`
(tho it reloads the client, could be from invalid url)

src/main/mainWindow.ts
```diff
@@ -455,18 +455,34 @@ function createMainWindow() {
...
 
     return win;
 }
 
 const runVencordMain = once(() => require(join(VENCORD_FILES_DIR, "vencordDesktopMain.js")));
 
+export function loadUrl(uri: string | undefined) {
+    const branch = Settings.store.discordBranch;
+    const subdomain = branch === "canary" || branch === "ptb" ? `${branch}.` : "";
+    mainWin.loadURL(`https://${subdomain}discord.com/${uri?.replace(RegExp("^discord://[^/]*/?"), "") || "app"}`);
```
Vendicated, line 475: you should use the URL api instead of regex for more robust handling

src/main/ipc.ts
```diff
@@ -135,6 +135,11 @@ handle(IpcEvents.CLIPBOARD_COPY_IMAGE, async (_, buf: ArrayBuffer, src: string)
     });
 });
 
+handle(IpcEvents.PAGE_NOT_FOUND_HANDLER, _ => {
+    loadUrl(undefined);
```
Vendicated, line 139: uh?

src/preload/index.ts
```diff
@@ -24,6 +24,10 @@ const style = document.createElement("style");
 style.id = "vcd-css-core";
 style.textContent = readFileSync(rendererCss, "utf-8");
 
+document.addEventListener("DOMContentLoaded", () => {
+    if (document.title === "Page Not Found | Discord") ipcRenderer.invoke(IpcEvents.PAGE_NOT_FOUND_HANDLER);
```
Vendicated, line 28: this is incredibly hacky. what if your discord is not in english?

src/main/mainWindow.ts
```diff
@@ -455,18 +455,35 @@ function createMainWindow() {
...
 
-    const subdomain =
-        Settings.store.discordBranch === "canary" || Settings.store.discordBranch === "ptb"
-            ? `${Settings.store.discordBranch}.`
-            : "";
+    let uriFiredDarwin = false;
+    app.on("open-url", (_, url) => {
+        if (uriFiredDarwin) restoreVesktop();
+        else loadUrl(url);
+        uriFiredDarwin = true;
```
Vendicated, line 462: why is this darwin only code running on windows and linux too?

Vendicated: this also doesn't look right

Vendicated: > As with the open-file event, be sure to register a listener for the open-url event early in your application startup to detect if the application is being opened to handle a URL. If you register the listener in response to a ready event, you'll miss URLs that trigger the launch of your application.

verysillycat:
This doesn't seem to work on Arch Linux, atleast not for me.

Arcitec:
Looks clean for the most part.

What's up with `uriFiredDarwin`? The logic does not make sense (never calling `loadUrl` if the value is true, and calling it **twice** if the value is false):

```js
    app.on("open-url", (_, url) => {
        if (uriFiredDarwin) restoreVesktop();
        else loadUrl(url);

...cut...

    if (!uriFiredDarwin) loadUrl(uri);
```

and there's no mention of needing to do that in the official Electron docs:

https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app

Vendicated:
did anyone test this with flatpak?

src/main/mainWindow.ts
```diff
@@ -499,5 +502,13 @@ export async function createWindows() {
         });
     });
 
+    mainWin.webContents.on("did-navigate", (_, url: string, responseCode: number) => {
```
Vendicated, line 505: why is this necessary?

Arcitec:
I've only had a really quick look at the new commit but the code definitely makes a lot more sense now. :) Now I finally understand its purpose.

So it seems like when Vesktop is being called as URL Open handler, on Mac it triggers an "open-url" event, and on Linux/Windows it instead triggers a "app executed with arg `discord://theurl`".

The new code makes sense and seems robust. Nice work. :)

https://github.com/Vencord/Vesktop/pull/813/commits/4464752a81fbe5ade1e2bff29b18a2db8bfe3ac9

src/main/mainWindow.ts
Arcitec: Seems like it might somehow prevent an infinitely looping/refreshing webpage. But it's unclear why it is needed yeah.

The event triggers when the page URL changes (not just anchors but actually navigation to another page):

https://www.electronjs.org/docs/latest/api/web-contents#event-did-navigate

And if this code detects that the new page is not starting with `/app` and the HTTP status code was 300 or higher (errors), it sets the URL to `undefined` to kill the browser instead of letting it loop on an error page.

That's as far as I got but not really sure *why* this is needed.

4A8:
does this not work with args? for example jumping to a message or opening invites
if not then it's possible to achieve the same thing using registry keys on windows and some commands on linux (and the args are not working for me)

4A8:
there are deeplinks that allow you to jump to a message that work on the normal discord app, but since arguments don't seem to be supported here they wont work
i think they look something like this: `discord://-/channels/<server-id>/<channel-id>/<message-id>`

Vendicated:
⚠️ This branch has conflicts that must be resolved
Use the [web editor](https://github.com/Vencord/Vesktop/pull/813/conflicts) or the  to resolve conflicts.
Conflicting files
src/main/mainWindow.ts

src/main/index.ts
```diff
@@ -112,6 +114,12 @@ async function bootstrap() {
     }
 }
 
+// MacOS only event
+export var darwinURL: string | undefined;
```
Vendicated, line 118: you will be sacrificed to the flying spaghetti monster soon if you don't change this var to a let
```suggestion
export let darwinURL: string | undefined;
```

src/main/mainWindow.ts
```diff
@@ -499,5 +502,13 @@ export async function createWindows() {
...
+        // check url to ensure app doesn't loop
+        if (new URL(url).pathname !== `/app` && responseCode >= 300) {
```
Vendicated, line 507:
```suggestion
        if (responseCode >= 300 && new URL(url).pathname !== `/app`) {
```

src/main/mainWindow.ts
```diff
@@ -499,5 +502,13 @@ export async function createWindows() {
...
+            loadUrl(undefined);
+            console.warn(`Caught bad page response: ${responseCode}, dumping to main app`);
```
Vendicated, line 509:
```suggestion
            console.warn(`'did-navigate': Caught bad page response: ${responseCode}, redirecting to main app`);
```

Vendicated:
awesome sauce

WiLuX-Source:
Is there any way to disable this? disabling desktop_deeplink experiment doesn't work.

Vendicated:
use our support channel instead of necroing closed prs.

Issue #790, opened 2024-07-30
Automatic Streamer Mode for OBS


Krutonium:
### Plugin Description

I'd love it if Streamer Mode on Linux could be automatically toggled by `obs-studio` being running - Right now it's not.

It would also benefit from a way to turn off or customize how much a name is truncated when in streamer mode.

### Request Agreement

- [X] I read the above and made sure my plugin is possible and allowed
- [X] I checked the [existing plugins](https://vencord.dev/plugins) and made sure my plugin doesn't already exist
- [X] I [searched the existing issues](https://github.com/Vencord/plugin-requests/issues?q=is%3Aissue) and made sure a similar request doesn't already exist

PR #822, opened 2024-08-20
arrpc: support OBS auto streamer mode

Covkie:
resolves: #790 

note: as usual with arrpc process scanning— this will not work in the flatpak. see https://github.com/flatpak/flatpak/issues/3922

### DIFF
--- a/patches/arrpc@3.5.0.patch
+++ b/patches/arrpc@3.5.0.patch
@@ -1,14 +1,31 @@
 diff --git a/src/process/index.js b/src/process/index.js
-index 97ea6514b54dd9c5df588c78f0397d31ab5f882a..c2bdbd6aaa5611bc6ff1d993beeb380b1f5ec575 100644
+index 389b084..d878301 100644
 --- a/src/process/index.js
 +++ b/src/process/index.js
-@@ -5,8 +5,7 @@ import fs from 'node:fs';
- import { dirname, join } from 'path';
- import { fileURLToPath } from 'url';
+@@ -1,12 +1,20 @@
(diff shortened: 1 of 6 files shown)


Vendicated:
i get the idea but this seems extremely hacky and i don't think having such a big patch is good for maintainability

at this point we just need a fork / our own thing

src/renderer/arrpc.ts
```diff
@@ -19,8 +19,14 @@ onIpcCommand(IpcCommands.RPC_ACTIVITY, async data => {
     if (!Settings.store.arRPC) return;
 
     await onceReady;
-
-    arRPC.handleEvent(new MessageEvent("message", { data }));
+    const StreamerModeStore = findStore("StreamerModeStore");
```
Vendicated, line 22: This is very bad. This find should be moved to the top level with findStoreLazy

patches/arrpc@3.5.0.patch
```diff
@@ -1,14 +1,31 @@
...
- import { dirname, join } from 'path';
- import { fileURLToPath } from 'url';
+@@ -1,12 +1,20 @@
+ const rgb = (r, g, b, msg) => `\x1b[38;2;${r};${g};${b}m${msg}\x1b[0m`;
+ const log = (...args) => console.log(`[${rgb(88, 101, 242, 'arRPC')} > ${rgb(237, 66, 69, 'process')}]`, ...args);
  
+-import fs from 'node:fs';
+-import { dirname, join } from 'path';
+-import { fileURLToPath } from 'url';
+-
```
Vendicated, line 12: removing these imports makes the diff unnecessarily complicated. just leave them in

PR #826, opened 2024-08-23
fix(rpm): don't generate build_id links to avoid conflicts

donCESAR12345:
On RHEL based systems a fix must be made to Electron apps to not include debugging symbols (see similar issue for vscode repository [here](https://github.com/microsoft/vscode/pull/116105/)).

This PR adds the fix to the package.json in order to build rpms without those debugging symbols and be able to install multiple electron apps without issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -179,6 +179,11 @@
         },
         "publish": {
             "provider": "github"
+        },
+        "rpm": {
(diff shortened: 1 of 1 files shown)


Vendicated:
thanks a lot for the fix!

Issue #161, opened 2023-10-23
Automatic Gain Control toggle


Vendicated:
Discord has a toggle for automatic gain control. It currently has no effect on vesktop though. it would be great to implement it

Plarpoon:
Hi all, wanted to report that the problem described in issue #142 is still present in the current version!

I myself didn't realize it was Vesktop bug so I actually spent a day looking at pavucontrol and banging my head against it. Thank you for reporting that it cannot be disabled, you put my mind at peace!

PopLamina:
This really needs to be fixed, very infuriating.

makidoll:
It might be a little cursed but I cloned the project and added the source code from this chrome extension into `src/renderer/index.ts`. It patches `getUserMedia` and such to permanently disable agc.

https://chromewebstore.google.com/detail/disable-automatic-gain-co/clpapnmmlmecieknddelobgikompchkk

jessicamaybe:
> It might be a little cursed but I cloned the project locally and added the source code from this chrome extension into `src/renderer/index.ts`. It patches `getUserMedia` and such to permanently disable agc.
> 
> https://chromewebstore.google.com/detail/disable-automatic-gain-co/clpapnmmlmecieknddelobgikompchkk

Would you mind sharing your changes if possible?

makidoll:
```diff
diff --git a/src/renderer/index.ts b/src/renderer/index.ts
index ebe6bc6..d891c26 100644
--- a/src/renderer/index.ts
+++ b/src/renderer/index.ts
@@ -57,3 +57,81 @@ VesktopNative.arrpc.onActivity(data => {
 
     arRPC.handleEvent(new MessageEvent("message", { data }));
 });
+
+// force disable automatic gain control
+
+(function () {
+    function setLegacyChromeConstraint(constraint, name, value) {
+        if (constraint.mandatory && name in constraint.mandatory) {
+            constraint.mandatory[name] = value;
+            return;
+        }
+        if (constraint.optional) {
+            const element = constraint.optional.find(opt => name in opt);
+            if (element) {
+                element[name] = value;
+                return;
+            }
+        }
+        // `mandatory` options throw errors for unknown keys, so avoid that by
+        // setting it under optional.
+        if (!constraint.optional) {
+            constraint.optional = [];
+        }
+        constraint.optional.push({ [name]: value });
+    }
+    function setConstraint(constraint, name, value) {
+        if (constraint.advanced) {
[...]

PR #832, opened 2024-08-31
Force disable automatic gain control

Justus0405:
This code forcefully disables automatic gain control, a feature that adjusts microphone volume system-wide, even in Windows when apps are denied permission to change microphone settings. The code was developed by Joey Watts and is licensed under the MIT License, I've been using it for two weeks, and it works perfectly.


Justus0405:
fixes #161

Justus0405:
Also confirmed to be an issue on Linux

Vendicated:
it's not a good idea to **force** disable something that people might want enabled. it should be tied to the autogain switch discord has

Novantric:
Make it a plugin?

thibaultmol:
Please make this an easy-to-enable thing.
It's so frustrating when your friends ask every five minutes why you're so quiet (and it turns out that it was adjusting your mic volumes once again)

PR #874, opened 2024-09-24
Enable speech-dispatcher support for TTS on linux

adryd325:


adryd325:
Have not tested if this has any adverse-effects on windows or macos

Vendicated:
thanks! I didn't know it's possible to pass that flag via electron api. I wonder if there are any drawbacks... surely it would be enabled by default if it didn't

Vendicated:
well only one way to find out

PR #929, opened 2024-10-26
feat: setWindowButtonPosition based on zoom factor for macOS

darwinx64:
https://github.com/user-attachments/assets/19b862f0-6a64-4ef2-a058-cc6fe1c97d4a

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -154,6 +154,15 @@ async function clearData(win: BrowserWindow) {
 
 type MenuItemList = Array<MenuItemConstructorOptions | false>;
 
+function setTrafficLightPositions(win: BrowserWindow) {
+    const horizontalOffset = (7 * win.webContents.zoomLevel) + 10;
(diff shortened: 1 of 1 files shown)


src/main/mainWindow.ts
```diff
@@ -235,14 +244,67 @@ function initMenuBar(win: BrowserWindow) {
             click() {
                 app.quit();
             }
+        }
+    ] satisfies MenuItemList;
+    const viewSubMenu = [
```
Vendicated, line 249: whats this

darwinx64: What u mean

Vendicated: why are you making a new menu unrelated to the feature here

darwinx64: This is the only way of detecting zoom changing as far as I know
I believe there's like window.webContents.on("zoom-change"), but it just doesn't work at all or it's for something else like trackpad zooming but im not sure 
You could detect Cmd + plus/minus but this wouldn't detect when zoom is changed from menubar but if you add a new menu it detects both that and change in menubar so it seems to be the all around best solution

PR #941, opened 2024-11-01
Autofix: [Bug] There's too much windows

lattaai1:
I've identified the issue causing multiple windows to open on launch. The problem appears to be in the `createWindows` function in `src/main/mainWindow.ts`. I've modified this function to ensure only one main window is created and to handle the case of an existing window properly.

Here's a summary of the changes:

1. Added a check to see if a main window already exists before creating a new one.
2. If a main window exists, we now focus and show it instead of creating a new one.
3. Added error handling to prevent crashes if window creation fails.

These changes should resolve the issue of multiple windows opening on launch while maintaining the existing functionality. 
> [!CAUTION]  
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission


verysillycat:
💀

PR #1011, opened 2024-12-16
Match installer requirements with Vesktop's requirements

### LINKED ISSUE #1001: [Bug] Installer works on Windows 7, despite Vesktop requiring Windows 10+
### Discord Account

korbosoft

### Operating System

Windows 7

_No response_

### Package Type

Installer

Despite Vesktop requiring Windows 10+, the installer asks for Windows 7+ instead.

Vesktop's installer should probably check for Windows 10 and up, to match with Vesktop's requirements (so you can't install Vesktop on a version too old for it to even run)

1. Install Vesktop, installation will succeed
2. Attempt to run Vesktop
3. An error will appear due to missing system functions

### Debug Logs

Don't think the installer would have anything remotely relevant, as it installs successfully (when it shouldn't).

### Request Agreement

korbosoft:
Adds a few lines to `installer.nsh` to check for at least Windows 10. fixes #1001. nin0 said I should PR anyways

### DIFF
--- a/build/installer.nsh
+++ b/build/installer.nsh
@@ -1,5 +1,10 @@
+!include WinVer.nsh
 !macro preInit
- SetRegView 64
+${IfNot} ${AtLeastWin10}
+  MessageBox mb_iconStop "Vesktop requires at least Windows 10."
+  Abort
+${EndIf}
+SetRegView 64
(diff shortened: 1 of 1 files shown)


korbosoft:
i'll repurpose this as a soft warning later

korbosoft:
or not lol

PR #1016, opened 2024-12-19
feat(rpc): implement arrpc's `link` event

Covkie:
This implements deep links from web so message links and the like work over just invites working.

Depends on OpenAsar/arrpc/pull/128

I noticed we didn't implement this after double checking what I said in a #813 comment

note: you need to be signed out of web for discord to prompt you for this

### DIFF
--- a/src/main/arrpc.ts
+++ b/src/main/arrpc.ts
@@ -26,8 +26,8 @@ export async function initArRPC() {
 
             await sendRendererCommand(IpcCommands.RPC_INVITE, invite).then(callback);
         });
-        server.on("link", (data: any, deepCallback: (valid: boolean) => void) => {
-            mainWin.webContents.executeJavaScript(`Vesktop.openDeepLink(${JSON.stringify(data)})`).then(deepCallback);
+        server.on("link", async (data: any, deepCallback: (valid: boolean) => void) => {
+            await sendRendererCommand(IpcCommands.RPC_DEEP_LINK, data).then(deepCallback);
         });
     } catch (e) {
         console.error("Failed to start arRPC server", e);
--- a/src/main/arrpc.ts
+++ b/src/main/arrpc.ts
@@ -26,6 +26,9 @@ export async function initArRPC() {
 
             await sendRendererCommand(IpcCommands.RPC_INVITE, invite).then(callback);
         });
+        server.on("link", (data: any, deepCallback: (valid: boolean) => void) => {
+            mainWin.webContents.executeJavaScript(`Vesktop.openDeepLink(${JSON.stringify(data)})`).then(deepCallback);
+        });
     } catch (e) {
         console.error("Failed to start arRPC server", e);
     }
--- a/src/renderer/arrpc.ts
+++ b/src/renderer/arrpc.ts
@@ -45,7 +45,8 @@ onIpcCommand(IpcCommands.RPC_DEEP_LINK, async data => {
     try {
         DEEP_LINK.handler({ args: data });
         return true;
-    } catch {
+    } catch (err) {
+        console.error("[RPC]", "Failed to open deep link:", err, data);
         return false;
     }
 });
--- a/src/renderer/arrpc.ts
+++ b/src/renderer/arrpc.ts
@@ -5,7 +5,7 @@
  */
 
 import { onceReady } from "@vencord/types/webpack";
-import { FluxDispatcher, InviteActions } from "@vencord/types/webpack/common";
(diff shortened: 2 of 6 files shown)


4A8:
this still doesn't work.. it still only switches to the window when i open a discord:// link that is supposed to jump to a message, but when i open a message link in the browser and then press open in discord it doesnt do anything

src/main/arrpc.ts
```diff
@@ -30,6 +30,9 @@ export async function initArRPC() {
                 .executeJavaScript(`Vesktop.openInviteModal(${JSON.stringify(invite)})`)
                 .then(callback);
         });
+        server.on("link", (data: any, deep_callback: (valid: boolean) => void) => {
```
Vendicated, line 33: js / vencord use camelCase

src/main/arrpc.ts
```diff
@@ -30,6 +30,9 @@ export async function initArRPC() {
...
+            mainWin.webContents.executeJavaScript(`Vesktop.openDeepLink(${JSON.stringify(data)})`).then(deep_callback);
```
Vendicated, line 34: i know you copy pasted this from the invite but I will add proper IPC that isn't just eval to main and then you can switch ur PR to that

src/renderer/arrpc.ts
```diff
@@ -38,3 +38,20 @@ onIpcCommand(IpcCommands.RPC_INVITE, async code => {
 
     return true;
 });
+
+onIpcCommand(IpcCommands.RPC_DEEP_LINK, async data => {
+    // console.log(data);
```
Vendicated, line 43:
```suggestion
```

src/renderer/arrpc.ts
```diff
@@ -38,3 +38,20 @@ onIpcCommand(IpcCommands.RPC_INVITE, async code => {
...
+    if (data.type === "CHANNEL") {
+        // I am unaware of any other types but ensure just in case.
```
Vendicated, line 45: this comment seems in the wrong place? should just be removed anyway

src/renderer/arrpc.ts
```diff
@@ -38,3 +38,20 @@ onIpcCommand(IpcCommands.RPC_INVITE, async code => {
...
+        const { guildId, channelId, messageId } = data.params;
+        if (!guildId) return false; // ensure at least guildId exists
+
+        const path = [guildId, channelId, messageId].filter(Boolean).join("/");
+        NavigationRouter.transitionTo(`/channels/${path}`);
```
Vendicated, line 50:
```suggestion
        NavigationRouter.transitionToGuild(guildId, channelId, messageId);
```

src/renderer/arrpc.ts
```diff
@@ -38,3 +38,14 @@ onIpcCommand(IpcCommands.RPC_INVITE, async code => {
...
     return true;
 });
+
+const { DEEP_LINK } = findLazy(m => m.DEEP_LINK?.handler);
+
+onIpcCommand(IpcCommands.RPC_DEEP_LINK, async data => {
+    try {
+        DEEP_LINK.handler({ args: data });
+        return true;
+    } catch {
```
Vendicated, line 48: this error should be caught and logged to make debugging easier

src/renderer/patches/nativeFocus.tsx
```diff
@@ -0,0 +1,23 @@
...
+addPatch({
+    patches: [
+        {
+            find: ".DEEP_LINK]:{",
+            replacement: [
+                {
+                    // TODO: Fix eslint rule
+                    // eslint-disable-next-line no-useless-escape
+                    match: /switch\(\i.\i.focus\(\)/,
+                    replace: "switch(VesktopNative.win.focus()"
```
Vendicated, line 18: I wouldn't rely on the weird switch code structure here too much since the bundler could very well change it

```suggestion
                    match: /(?<=\.DEEP_LINK.{0,200}?)\i\.\i\.focus\(\)/,
                    replace: "VesktopNative.win.focus()"
```

src/shared/IpcEvents.ts
```diff
@@ -61,6 +61,7 @@ export const enum IpcEvents {
 export const enum IpcCommands {
     RPC_ACTIVITY = "rpc:activity",
     RPC_INVITE = "rpc:invite",
+    RPC_DEEP_LINK = "rpc:deep",
```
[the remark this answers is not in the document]
Vendicated: im 14 and this is deep

Issue #1009, opened 2024-12-15
[Bug] Vesktop lag/stutter on Wayland with Rich Presence + ozone-platform-hint=auto


Kitt3120:
### Discord Account

schweren.dev

### Operating System

NixOS

### Linux Only ~ Desktop Environment

KDE Plasma 6.2.4 on Wayland 

### Package Type

nixpkg, AppImage

### What happens when the bug or crash occurs?

When using the `ozone-platform-hint=auto` command line argument, and having the Rich Presence feature enabled, Vesktop lags/stutters about every 5 seconds.

Here's a video demonstrating this: https://www.youtube.com/watch?v=YRhVis_xId8

What I found out so far:

- The lags/stutters happen only when `ozone-platform-hint=auto` is set and the Rich Presence feature is enabled
- When disabling Rich Presence, it still lags as long as Vesktop hasn't been restarted
- The lags/stutters happen around every 5 seconds, very consistently. The interval is **not** affected by system performance. It's about 5 seconds on my desktop and my laptop. But the duration of the lag depends on the system's performance. The lag is way shorter on my desktop than on my laptop.
- It does not have anything to do with 3rd party plugins. With all plugins disabled, it still happens.
- It does not have anything to do with CSS/Theme related settings. With custom css disabled, and Discord's default theme applied, it still happens.
- It does not have anything to do with GPU hardware acceleration. With hardware acceleration disabled in settings, and starting with the `disable-gpu` flag, it still happens.
- The lags can be observed visually, and in the browser devtools' "performance" section (ctrl + shift + i), like shown in the video. I recommend to record at least 10 seconds to see at least 2 lags. Lags can be seen in the performance result as "dropped frames".

My guess is that there's a 5-second timer related to the Rich Presence feature, running some task. Whatever that task does, it freezes the UI for a short moment.

### What is the expected behaviour?

Vesktop does not stutter about every 5 seconds

### How do you recreate this bug or crash?

1. Use a Wayland compositor (in my case KDE Plasma's Wayland session)
2. Run Vesktop with the `ozone-platform-hint=auto` flag
3. Enable Rich Presence (either have it enabled already or switch the setting from off to on after starting Vesktop)
4. Vesktop now lags/stutters about every 5 seconds. This is **not only when playing video** (like shown in the linked YouTube video), but also when scrolling through chats, typing a message, etc. Basically, it's periodically lagging, no matter what you're doing.

(Optional)
1. Disable Rich Presence in settings
[...]

PR #1053, opened 2025-01-16
feat: move arrpc server into a worker thread

pendo324:
Solution for #1009, which I can reliably reproduce without this change.

I have another version of this which instead works with this arrpc PR https://github.com/OpenAsar/arrpc/pull/133, but it seems like Vesktop might be the appropriate place to put this complexity.

I have tested that game status RPC works as expected but I haven't found a game with which to test invites.


pendo324:
Off by one error, works on my side now

pendo324:
> Instead of writing an insane review with github web editor heres a patch. It's functionally the same but doesn't have the hardcoded types (?) or add a new dep
> 
> [buh.patch](https://github.com/user-attachments/files/18432602/buh.txt)

Thanks. I removed the new dependency in my latest revision. Not sure why we wouldn't want the types?

pendo324:
> You're over complicating it by quite a bit

Agree to disagree, I suppose. It's much more readable to me with the MessageChannel, and adds like... 1 line of code, which is why I didn't use `postMessage`. I’m surprised your code works using a `.ts` file for the worker, even when it isn’t explicitly loaded using an import/require (you reference the js output, which I’m surprised is included by ESBuild in the bundle when it’s not ever imported). Let me try removing the new entrypoint I added and see if that works.

You also use a map instead of an array for storing the invites, which is fine too, but I don’t really see why you’d need an id more unique than a monotonically increasing array index for that. Maybe it’s more extendable.

> Also please stop forcing pushing it makes it insane to follow changes :sob: 

Sure. Habit from other projects I contribute to where this is the expected workflow. I just use the "see changes since last revision" button.

Vendicated:
Thank you for your PR!

I took the freedom to refactor it to be more idiomatic and use more appropriate data structures (Map with uuids over an Array with indexes)

Besides that it looks good

Vendicated:
just realised you forgot to handle the link event inside the worker

Vendicated:
https://github.com/Vencord/Vesktop/commit/4baf6de4729c3cf389376c892068ec065931b05b

PR #1060, opened 2025-01-24
feat: preferred stream settings

antoniosarro:
Save stream settings for reuse (resolution, FPS, content type). Audio source is excluded due to per-stream variability.


Vendicated:
please show a screenshot

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -453,6 +453,25 @@ function StreamSettings({
...
             </Card>
+
+            <div className="vcd-screen-picker-save-settings">
+                <label htmlFor="" className="vcd-screen-picker-save-label">
+                    Remember stream settings
+                </label>
+                <div className="vcd-screen-picker-save-control">
+                    <div
+                        className={`vcd-screen-picker-save-checkbox ${Settings.stream?.preferred && "checked"}`}
+                        onClick={() => {
```
Vendicated, line 464: this markup is very messy and not semantic (using a plain div as a checkbox is very bad. at the very least it should have aria attributes like `role` that tell user agents that this is a checkbox for visually impaired people and other users of screenreaders). have a look at the rest of Vesktop (& Vencord) Settings UI and see how those implement things like Switches, Checkboxes. Vencord exports a bunch of Discord Components you can use instead of implementing stuff yourself

antoniosarro:
image

I initially attempted to use the existing Switch component, but it appeared inconsistent with the current UI due to its hardcoded colors. I'm open to using it if it's the preferred solution. In that case, I can refine the markup based on your suggestions.

Vendicated:
I see, that's fair yeah! Switch is likely not the right Component, but Discord also has a Checkbox component that should be exactly what you want

image

Vencord doesn't export it for some reason, but you can manually find it via `const CheckBox = findExportedComponentLazy("Checkbox", "Switch");`. Search (Ctrl + Shift + F) for `.Checkbox` in sources in chrome devtools to find how Discord uses it & which arguments you can pass

image

antoniosarro:
I updated the component using the Checkbox component.

Regarding preferred settings for audio, currently, only the "Entire System" option can be saved. However, I believe it's best to exclude audio settings altogether.

image

Vendicated:
Thanks for your contribution! However, I added an alternative approach for this in https://github.com/Vencord/Vesktop/commit/b620e07445a672e556255d1df200bf1b777c4f8e that just always remembers resolution & fps

Issue #18, opened 2023-04-23
Custom Keybinds


TheKodeToad:
![image](https://user-images.githubusercontent.com/57493648/233842841-5780a481-be60-4b15-91e5-046bc080f4b6.png)

makidoll:
I really love Vencord, it's so nice to use including the better screen sharing support on Linux. But this is the only feature I really wish it had. I wonder if it can be done with a plugin?

libjared:
I'm thinking of leveraging the new PluginNative functionality that was committed to Vencord a few days ago.
It'd expose any keybindable Discord action (but mainly PTT, initially) to the kglobalaccel DBus, allowing users to setup keybinds via KDE systemsettings.

D3SOX:
We already talked about this in the Discord and I think it would probably be better to just implement this in Vesktop using [Electron global shortcuts API](https://www.electronjs.org/docs/latest/api/global-shortcut) as this is already cross-platform

Vendicated:
yes, this should definitely be part of vesktop and not a plugin

either via electron's global shortcut api, or via [ihook](https://github.com/wilix-team/iohook)

the former does not support things like mouse side buttons and has no way to implement PTT, the latter would require a native module

Curve:
> yes, this should definitely be part of vesktop and not a plugin
> 
> either via electron's global shortcut api, or via [iohook](https://github.com/wilix-team/iohook)
> 
> the former does not support things like mouse side buttons and has no way to implement PTT, the latter would require a native module

Correct me if I'm wrong but as the latter uses `libuiohook` I think that it will not support wayland.

I think the electron shortcut API is the right way to go as it is probably the most portable and clean

RinLovesYou:
I am working on this. iohook is old and unmaintained. I am using uiohook-naapi to do this. it also uses libuiohook, and it seems wayland support is in the works. Nothing you can do but wait until libuiohook updates for that, but i'll go ahead and keep using it.

Curve:
> I am working on this. iohook is old and unmaintained. I am using uiohook-naapi to do this. it also uses libuiohook, and it seems wayland support is in the works. Nothing you can do but wait until libuiohook updates for that, but i'll go ahead and keep using it.

Why not use the xdg-desktop-shortcuts? It's way more clean & portable imo (no low level hooking and such)

RinLovesYou:
Didn't know about those, I'll look into it

RinLovesYou:
[...]

PR #1072, opened 2025-02-01
Add D-Bus interface

notpeelz:
I made a simple D-Bus interface to interact with Discord. Currently, it has 2 methods (`ToggleMute` and `ToggleDeafen`) and 2 properties (`Muted`, `Deafened`).

https://github.com/user-attachments/assets/9a177d8b-c07d-4a6c-afe8-7386f987df9d

I'm aware there's a PR in the works for global shortcuts, which makes use of the xdg-desktop-portal GlobalShortcuts API (#326). Personally, I don't care much for the GlobalShortcuts API, as it doesn't work well with declaratively configured compositors (e.g. sway, niri, etc), as well as many niche compositors that may never implement GlobalShortcuts. I don't intend for this D-Bus API to replace global shortcuts.

The main motivation for exposing a D-Bus interface is to allow external programs to interact with Discord, such as adding a mute/deafen button to a VR overlay or displaying the muted/deafened state on your desktop shell or VR environment.

I implemented the D-Bus code as a native module in a separate repo, but it could probably live as a subfolder in the Vesktop repo (I followed the same structure as venmic): https://github.com/notpeelz/venbus

Feedback welcome.

Related: #18


notpeelz:
A D-Bus API isn't a replacement for global shortcuts, it's orthogonal. It would allow for your desktop shell (or VR overlay) to display the muted/deafened status (and could be expanded to expose even more information, such as the current voice channel, number of users in VC, etc... you can use your imagination).

While #326 has a way to invoke shortcuts via CLI, it requires booting up an entire instance of electron just to do IPC... not to mention that `vesktop --keybind 2` isn't exactly very descriptive or discoverable.

I asked around on the LVRA (Linux VR Adventures) discord and the feedback was rather positive for a feature like this.
Is there room for discussion?

Sqaaakoi:
I feel like D-Bus would be a better solution for desktop shells in particular. I've actually tried (probably about a year and a half ago) to make something like this, and failed because of how bad Discord's own RPC protocol is. I actually tried to make my own protocol through a plugin that listened for connections on a unix socket as a result of that.

I don't think brushing off D-Bus because it's overcomplicated is the right solution, especially not considering that there is currently no alternative for this RPC functionality in Vesktop.
At the same time, I don't think merging this into Vesktop is the right call either; as D-Bus functionality inside of Vencord plugins is a use case that I would find useful, that being richer desktop notification support. As such, I feel like this would be generally more useful as a feature of Vencord itself.
I am aware that this idea is crazy. I can't promise that this will be added to Vencord, so I would recommend implementing it as a user plugin for now (unless the native functionality available inside of a plugin directory isn't enough to make that work).

notpeelz:
I agree with @Covkie that arrpc would be the right way to implement more advanced features (like muted/deafened status, etc). Though, on Linux, D-Bus is ubiquitous, making it the natural choice for IPC.
Reimplementing all of discord's RPC and exposing it via D-Bus doesn't seem like the right way to go, but surely there's a middle ground somewhere.

I can give userplugins a try, but I don't really fancy the idea of maintaining a fork just for this... if only there was a way to dynamically load plugins runtime :thinking:

notpeelz:
I don't think I can do this in a user plugin since I need to load a native module, which can only be done from the main process. From what I understand, (user)plugins only run in the renderer process.

Issue #1077, opened 2025-02-04
[Feature Request] Add Window Controls to new Discord Title Bar


Keqwerty:
### Discord Account

keqwerty

### Motivation

So Discord just released a new UI without any choice. There are no window control buttons on Linux when I use Discord's title bar (I can't disable it because the default frame window system on Wayland works weird)
https://i.imgur.com/q7A4EJn.png

### Solution

Just add new window control buttons to Discord Title Bar

### Alternatives

N/A

### Request Agreement

- [x] I have searched the existing issues and found no similar issue
- [x] This is not a plugin request
- [x] This is not a Vencord feature request

Keqwerty:
> for now you can unenroll yourself from the visual refresh experiment by enabling the experiments plugin & setting yourself to 'Not Eligible' for the "Desktop Visual Refresh" experiment. Use the support channel if you need more help in doing so.

wow, thanks!

Vendicated:
This will likely not be fixed until this redesign gets fully rolled out and becomes stable. right now it's only an in development experiment Discord is doing, so it is still subject to frequent change

KibSquib48:
> This will likely not be fixed until this redesign gets fully rolled out and becomes stable. right now it's only an in development experiment Discord is doing, so it is still subject to frequent change

Seems like it's rolling out widely starting today

nac0n:
Just got the update myself and I have the same issue. It's too big...

Let us (the users) customize if you're going to TOS-force us to one client.

senposage:
redesign is rolling and and broke the title bar

Vendicated:
yes we know. this will be fixed soon, just have patience

p0358:
To anyone implementing this, please remember to also ensure the buttons are shown in the settings (as the primary UI's main titlebar isn't shown there), there seems to be some free space above the scrollbars:

![Image](https://github.com/user-attachments/assets/be10c090-e64d-41c7-9c56-1a459f271334)

On Discord mockups it'd show them like this:

![](https://static.beebom.com/wp-content/uploads/2025/03/Discord-Desktop-Gets-a-Makeover-with-New-Themes-In-Game-Overlay-and-More.jpg)

alper-han:
It's really annoying

Issue #964, opened 2024-11-20
[Bug] Discord Livestream in a pop-out can't be minimized via Titlebar. Main Vekstop app gets minimized instead.


Zimnyte:
### Discord Account

Zimnyte

### Operating System

Windows 11 24H2 26100.1742

### Package Type

Setup exe

### What happens when the bug or crash occurs?

Just today switched to Vesktop from regular discord with Vencord installed. After opening a livestream in a pop-out window I quickly realized that minimize button in the titlebar is not minimizing the stream window but instead it minimizes the main discord app on my other monitor. Trying to minimize it via taskbar at the bottom by clicking works as intended.

### What is the expected behaviour?

Pop-out stream window should minimize instead of the main Vesktop app.

### How do you recreate this bug or crash?

1. Join in a call with someone streaming
2. Join the stream
3. Open it in a pop-out window
4. Try minimizing it with a Titlerbar
5. Main App gets closed instead

### Debug Logs

```
Replace this text with your crash-log. Do not remove the backticks
```

### Request Agreement

- [X] I have searched the existing issues and found no similar issue
- [X] I am using the latest Vesktop and Vencord versions
- [X] This issue occurs on an official release (not just the AUR or Nix packages)

PR #1104, opened 2025-02-18
Fix window controls (popouts and visual refresh) and focus window method by patching window controls wrapper module

Sqaaakoi:
Fixes https://github.com/Vencord/Vesktop/issues/1077
Fixes https://github.com/Vencord/Vesktop/issues/964

This patches Discord's module that wraps around native IPC methods for windowing and some other utilities to call `VesktopNative.win` methods instead of `DiscordNative.window` methods
This includes `focus` which was used by RPC deep linking. No difference will occur from this change, but if any of those other handlers are later supported through arRPC, it won't be necessary to patch the focus method in those handlers.

This does not fix the missing always on top support in call popouts (doesnt work on wayland, I gave up on fixing that for that reason)


src/main/ipc.ts
```diff
@@ -68,28 +68,32 @@ handle(IpcEvents.SHOW_ITEM_IN_FOLDER, (_, path) => {
     shell.showItemInFolder(path);
 });
 
+function getWindow(e, key?: string) {
```
Vendicated, line 71:
```suggestion
function getWindow(e: IpcMainInvokeEvent, key?: string) {
```

src/main/ipc.ts
```diff
@@ -68,28 +68,32 @@ handle(IpcEvents.SHOW_ITEM_IN_FOLDER, (_, path) => {
...
+    const win = BrowserWindow.fromWebContents(e.sender) ?? mainWin;
+    if (!key) return win;
+    const popout = PopoutWindows.get(key!);
+    return popout ?? win;
```
Vendicated, line 75:
```suggestion
	return key
		? PopoutWindows.get(key)
		: BrowserWindow.fromWebContents(e.sender) ?? mainWin;
```

Vendicated:
do not remove the old patches. not everyone is using Visual Refresh

Vendicated:
we should hide this discord titlebar if you have discord titlebar disabled

you won't be able to access your inbox but welp blame discord for being braindead and not testing their web app

image

Vendicated:
they will probably fix it soon

Vendicated:
looks like shit and expecting people to fix shit with css is bad UX

Issue #1107, opened 2025-02-20
[Bug] Weird colors when screensharing with hardware acceleration enabled on vesktop settings.


RalL64:
### Operating System

CachyOS x86_64 (Linux 6.13.3-2-cachyos kernel)

### Linux Only ~ Desktop Environment

KDE Wayland

### Package Type

Flatpak

### What happens when the bug or crash occurs?

Weird colors when screensharing, like screen artifacts and this only affects who watches the stream and not my preview. 

### What is the expected behaviour?

"Normal colors", 60FPS, any resolution and hardware acceleration working as it should. 

### How do you recreate this bug or crash?

1- Enable hardware acceleration on vesktop settings (not the discord one located in voice and video) 
2- Start screensharing the whole desktop and not a window
3- Stream will remain normal for a few seconds, colors will start to bug and glitch just like an GPU artifact
4- User preview is normal but for the people who watches the stream is bugged
5- Turns off hardware acceleration from vesktop settings and everything works as it should.

### Debug Logs

```shell
Wayland socket is available, running natively on Wayland.
To disable, remove the --socket=wayland permission.
Passing the following arguments to Electron: --enable-speech-dispatcher --ozone-platform-hint=auto --enable-wayland-ime
[3:0220/093932.456687:ERROR:bus.cc(407)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
APPIMAGE env is not defined, current application is not an AppImage
Vesktop v1.5.5
checkForUpdatesAndNotify called, downloadPromise is null
LaunchProcess: failed to execvp:
xdg-settings
[3:0220/093932.660552:ERROR:bus.cc(407)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[3:0220/093932.660584:ERROR:bus.cc(407)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory

(vesktop.bin:3): libnotify-WARNING **: 09:39:36.566: Running in confined mode, using Portal notifications. Some features and hints won't be supported
[3:0220/094045.531125:ERROR:shared_x_display.cc(39)] Unable to open display
'loop->recurse > 0' failed at ../src/pipewire/thread-loop.c:425 pw_thread_loop_wait()
[2025-02-20 09:40:47.763] [venmic] [info] [patchbay] (handle) found default metadata: 38
[...]

Issue #1004, opened 2024-12-13
[Bug] Hardware encoding not working properly linux


developer-iname:
### Operating System

Linux(Arch)

### Linux Only ~ Desktop Environment

Sway

### Package Type

pipewire wireplumber xdg-desktop-wlr xdg-desktop-portal mesa

### What happens when the bug or crash occurs?

Stream is unviewable

### What is the expected behaviour?

I expect discord to stream and be viewable for others.

Actual result:
Others see an infinite loading screen, or the stream is garbled mess of black and white with some visible screen elements.

On launching vesktop from the terminal i see this
ERROR:vaapi_video_decoder.cc(1225)] failed Initialize()ing the frame pool

Which I believe is the culprit, I have not found a solution, however I believe it may be that vesktop is fallbacking to software decoding. As if the person recieving the stream on the other end disables their hardware encoding, it fixes the issue.

Also, perfomance while streaming is degraded massively.

### How do you recreate this bug or crash?

1. Stream
2. Receiver sees errors
3. End Stream
4. Receiver turns off hardware acceleration
5. Stream
6. Receiver sees stream

### Debug Logs

```
ERROR:vaapi_video_decoder.cc(1225)] failed Initialize()ing the frame pool
```

### Request Agreement

- [X] I have searched the existing issues and found no similar issue
- [X] I am using the latest Vesktop and Vencord versions
- [X] This issue occurs on an official release (not just the AUR or Nix packages)

mazeratibooloo:
Also having this exact same issue, I've been having it for nearly a week now. On Arch / KDE Plasma.

Skeletonek:
Have the same issue as you but my has been closed (https://github.com/Vencord/Vesktop/issues/997)
I am having the same error in console regarding va-api drivers on my Arch Linux system (Mesa 24.3.1)

I've checked running Vesktop 1.5.4 Appimage through distrobox running Ubuntu Noble with Mesa 24.0.9 and the stream seems to be working properly there. Running natively on arch is causing the stream to be super dark for the viewers of official discord client

mazeratibooloo:
If they plan to close this one similarly, it'd be nice to have an explanation further than marking the issue with tags and then just closing it. If it's not fixable by Vesktop and it's an issue with something else, it'd be great to know where to post an issue up about it.

Kerko0:
[...]

Issue #851, opened 2024-09-12
[Bug] Chroma artifacts when screensharing 720p with HW acceleration enabled.


silikeite:
### Operating System

Arch Linux

### Linux Only ~ Desktop Environment

KDE Wayland

### Package Type

Flatpak

### What happens when the bug or crash occurs?

Chroma issues with stream. This is only visible from other participants' sides; it looks normal from the preview.

![image](https://github.com/user-attachments/assets/bb18ae4d-d6e4-472c-a499-f8ba7652df38)

### What is the expected behaviour?

Normal image just as if I stream on 480p with HW Acceleration enabled or on 720p with HW Acceleration disabled.

### How do you recreate this bug or crash?

1. Use Flatpak on Intel iGPU KDE Wayland (not sure how important this is.)
2. Start Vesktop with HW Acceleration enabled
3. Stream anything at 720p 30fps.
4. Have someone (not yourself) look at the stream.
5. They should see the glitched colors.

### Debug Logs

```
❯ flatpak run dev.vencord.Vesktop
Using NVIDIA on Wayland, disabling gpu sandbox
Wayland socket is available, running natively on Wayland.
To disable, remove the --socket=wayland permission.
Passing the following arguments to Electron: --enable-speech-dispatcher --disable-gpu-sandbox --ozone-platform-hint=auto
[3 zypak-helper] Wait found events, but sd-event found none
[3:0912/134909.907381:ERROR:bus.cc(407)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
APPIMAGE env is not defined, current application is not an AppImage
checkForUpdatesAndNotify called, downloadPromise is null
[3:0912/134910.121583:ERROR:bus.cc(407)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[3:0912/134910.121612:ERROR:bus.cc(407)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[3:0912/134925.845911:ERROR:shared_x_display.cc(39)] Unable to open display
'loop->recurse > 0' failed at ../src/pipewire/thread-loop.c:426 pw_thread_loop_wait()
[2024-09-12 13:49:33.233] [venmic] [info] [patchbay] (handle) found default metadata: 39
[...]

Issue #797, opened 2024-08-04
[Bug] Graphical corruption on screenshare with Iris Plus Graphics


probuster366:
### Discord Account

alfdemar

### Operating System

Nobara 40

### Linux Only ~ Desktop Environment

KDE PLASMA - Wayland

### Package Type

Flatpak

### What happens when the bug or crash occurs?

There are graphical corruptions on screenshare feed when screensharing with hardware acceleration on under wayland session. 

Only 720p feed is affected by this bug, 480p and 1080p doesn't have any graphical corruptions.

My Specs:
CPU: Intel i3-1005G1 (4) @ 3.400GHz 
GPU: Intel Iris Plus Graphics G1

It looks like this:
![image](https://github.com/user-attachments/assets/6f87bbb8-c869-4844-80b2-a1dd3aa66571)
![mAEY4Hb](https://github.com/user-attachments/assets/c5e61ff2-b133-4913-b934-cb33c28e6349)

### What is the expected behaviour?

I expect vesktop to screenshare without graphical corruptions

### How do you recreate this bug or crash?

1. Make sure hardware acceleration is enabled on affected PC
2. Join a voice channel and start a screenshare feed at 720p (480p and 1080p are not affected by this bug)
3. To observe the bug, someone else should join the same voice channel and start watching the screensharing feed
4. To work around the bug, the affected host should disable hardware acceleration and restart vesktop and the feed should be fixed afterwards

### Debug Logs

```
Passing the following arguments to Electron: --enable-speech-dispatcher --ozone-platform-hint=auto
[3:0804/211124.916901:ERROR:bus.cc(407)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
APPIMAGE env is not defined, current application is not an AppImage
checkForUpdatesAndNotify called, downloadPromise is null
[3:0804/211125.266832:ERROR:bus.cc(407)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[3:0804/211125.266879:ERROR:bus.cc(407)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[arRPC > ipc] listening at /run/user/1000/discord-ipc-0
[arRPC > websocket] listening on 6463
[arRPC > process] started
[3:0804/211137.423046:ERROR:shared_x_display.cc(39)] Unable to open display
[...]

PR #1117, opened 2025-03-01
feat: Add a toggle in the settings for video hardware acceleration

AngelCMHxD:
Adds a toggle for video hardware acceleration in the Vesktop Settings while keeping overall hardware acceleration turned on, since video hardware acceleration may cause screen-share issues in some systems.

Would add a workaround to issues like #1107, #1004, #851 and #797

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -31,14 +31,16 @@ process.env.VENCORD_USER_DATA_DIR = DATA_DIR;
 function init() {
     app.setAsDefaultProtocolClient("discord");
 
-    const { disableSmoothScroll, hardwareAcceleration } = Settings.store;
+    const { disableSmoothScroll, hardwareAcceleration, videoHardwareAcceleration } = Settings.store;
 
     const enabledFeatures = app.commandLine.getSwitchValue("enable-features").split(",");
     const disabledFeatures = app.commandLine.getSwitchValue("disable-features").split(",");
 
     if (hardwareAcceleration === false) {
         app.disableHardwareAcceleration();
-    } else {
+    }
+    
+    if (hardwareAcceleration && videoHardwareAcceleration) {
         enabledFeatures.push(
             "AcceleratedVideoDecodeLinuxGL",
             "AcceleratedVideoEncoder",
--- a/src/renderer/components/settings/Settings.tsx
+++ b/src/renderer/components/settings/Settings.tsx
@@ -38,6 +38,13 @@ const SettingsOptions: Record<string, Array<BooleanSetting | SettingsComponent>>
             title: "Hardware Acceleration",
             description: "Enable hardware acceleration",
             defaultValue: true
+        },
+        {
+            key: "videoHardwareAcceleration",
+            title: "Video Hardware Acceleration",
+            description: "Improves screensharing performance. May cause color glitches. Requires a full restart.",
+            defaultValue: true,
+            disabled: () => Settings.store.hardwareAcceleration === false
         }
     ],
     "User Interface": [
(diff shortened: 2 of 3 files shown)


AngelCMHxD:
Oh, that does seem like a better way to do it, I'll take a look. Thanks!

AngelCMHxD:
After some long trial and error (while also discovering the Patch Helper thing in setting _kinda_ late .-.) I did the patch thingy! It was fun to do, however I really couldn't think of a better way to apply it other than saving it in the `Settings.store` object when it is toggled and then check the value on the `src/main/index.ts` like before. If you can think of one you can tell me and I'll check it out. :P

twixcider:
> After some long trial and error (while also discovering the Patch Helper thing in setting _kinda_ late .-.) I did the patch thingy! It was fun to do, however I really couldn't think of a better way to apply the change other than saving the value in the `Settings.store` object whenever it is toggled and then check the value on the startup over at `src/main/index.ts` like before. If you can think of one you can tell me and I'll check it out. :P

The switch doesn't seems to do anything, even I turned the hardware acceleration switch off the screen artifact still appear

AngelCMHxD:
> The switch doesn't seems to do anything, even I turned the hardware acceleration switch off the screen artifact still appear

If disabling Hardware Acceleration under Vesktop Settings doesn't fix the artifacts, then this PR won't do so too. For some people the cause of artifacts when screen sharing is video hardware acceleration, this PR is made so those people can disable *only* video hardware acceleration while not disabling all hardware acceleration.

However video hardware acceleration isn't always the cause, and after browsing the issues posted here about artifacts I don't know what else could it be. In some setups it just works, in others it's hardware acceleration what causes it, in others the cause is unknown. This PR is just to add a workaround (better than disabling all hardware acceleration) that may work for some people, but it still isn't a fix. :/

(If disabling Hardware Acceleration under Vesktop Settings did fix it but the toggle under Voice & Video doesn't while using this PR then hit me up)

twixcider:
> > The switch doesn't seems to do anything, even I turned the hardware acceleration switch off the screen artifact still appear
> 
> If disabling Hardware Acceleration under Vesktop Settings doesn't fix the artifacts, then this PR won't do so too. For some people the cause of artifacts when screen sharing is video hardware acceleration, this PR is made so those people can disable _only_ video hardware acceleration while not disabling all hardware acceleration.
> 
> However video hardware acceleration isn't always the cause, and after browsing the issues posted here about artifacts I don't know what else could it be. In some setups it just works, in others it's hardware acceleration what causes it, in others the cause is unknown. This PR is just to add a workaround (better than disabling all hardware acceleration) that may work for some people, but it still isn't a fix. :/
> 
> (If disabling Hardware Acceleration under Vesktop Settings did fix it but the toggle under Voice & Video doesn't while using this PR then hit me up)

Yeah I meant it that way, switching off hardware acceleration under vesktop setting does fix the artifact, however under video section the switch doesn't fix the artifact. so my workaround right no is to disable hardware acceleration under vesktop setting which make my whole vesktop laggy.

AngelCMHxD:
> Yeah I meant it that way, btw I don't know how github work does the pr you mention push in to the official repo? I am using the official repo while commenting. switching off hardware acceleration under vesktop setting does fix the artifact, however under video section the switch doesn't fix the artifact. so my workaround right no is to disable hardware acceleration under vesktop setting which make my whole vesktop laggy.

In a pull request the code isn't on the main branch or repo and will only be pushed once (or if) it's accepted, in order to use it right now you would need to clone my fork and compile it yourself. An easier alternative to compile it yourself is to launch the normal Vesktop with this launch argument from a terminal or a shortcut: `--disable-features=AcceleratedVideoEncoder` and it should only disable video encoding hardware acceleration and keep the rest working.

twixcider:
> --disable-features=AcceleratedVideoEncoder

thank you I will try the argument

Vendicated:
ah i missed this, sorry

I added this a few weeks back via https://github.com/Vencord/Vesktop/commit/48e9aea47e23f1e4df2aeb09f15ed47053dfacb4

thank you though!

PR #1129, opened 2025-03-07
fix macOS shortcut for dev tools (Cmd+Shift+I but need to press Shift+Control+II)

Mousica-dev:


Vendicated:
why add all that

Mousica-dev:
followed a guide on stack overflow and it looked like that, didnt think much of it until now. (could be shortened)

Mousica-dev:
weirdly enough, with this cmd sift i keybind you need to press shift control i for it to work in the actual app

Vendicated:
I still don't understand this PR. The DevTools shortcut is `Command+Option+I`, **not** `Command+Shift+I`

https://developer.chrome.com/docs/devtools/shortcuts

Mousica-dev:
Even when trying to press Cmd option I, it doesn’t work; it does manage to close dev tools if you alrdy opened it via View -> Toggle Dev Tools. I chose Cmd Shift I because I came from Windows, and it’s muscle memory atp, doesn’t really matter since it’ll probably fix itself in the next update, but https://github.com/electron/electron/issues/45925 says it’s not necessarily an Electron issue and something that Vesktop does to interfere. Either way, when you’ll bump the Electron version for Vesktop, I think it’ll just fix itself, speculation of course.

Vendicated:
The issue you linked doesn't say anything and is a bad issue anyway. Use the electron template to reproduce this issue on a minimal app, then report it to electron with that and share all relevant information

Vendicated:
https://github.com/Vencord/Vesktop/pull/1141

PR #1130, opened 2025-03-17
fix: don't close app on Ctrl+W

### LINKED ISSUE #934: [Bug] Window can be closed with Ctrl+W
### Discord Account

_No response_

### Operating System

Arch Linux

KDE Plasma(Wayland)

### Package Type

AUR and Appimage, version 1.5.3

The window closes when I press Ctrl+W, but the official discord client does not.
It happened on any page, whether it's on the login or chat page.

Nothing will happen.
Blocking browser shortcut keys shouldn't be that hard, I guess.

1. open a Vesktop window.
2. press Ctrl+W.

### Debug Logs

No new logs have been added, which is a very small thing.

### Request Agreement

- [x] This issue occurs on an official release (not just the AUR or Nix packages)

rushiiMachine:
Matches stock client behavior. If people want this back, they should add a keybind to their WM instead, which makes it configurable.

Fixes #934

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -6,7 +6,7 @@
 
 import "./ipc";
 
-import { app, BrowserWindow, nativeTheme } from "electron";
+import { app, BrowserWindow, globalShortcut, nativeTheme } from "electron";
 import { autoUpdater } from "electron-updater";
 
 import { DATA_DIR } from "./constants";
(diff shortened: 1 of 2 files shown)


Vendicated:
this is a really bad way to implement this because globalShortcut is actually global

rushiiMachine:
:trollface:

Vendicated:
do we still want these?

image

rushiiMachine:
I think they're useless, removing would be fine

PR #1132, opened 2025-03-17
fix: show tray icon when launching with `--start-minimized`

rushiiMachine:
If launching Vesktop with the `--start-minimized` option, there is no indication that Vesktop is running when the tray icon option is disabled. This forces the tray icon until the window is first shown, at which point it is hidden again if the tray icon option is off.


Vendicated:
I disagree

If the user explicitly disables the tray, we should respect that. You don't need the tray, you can just launch Vesktop via shortcut. If it's already open in the background, it will show that instead of opening a second instance

ghost:
LGTM

PR #1133, opened 2025-03-18
feat: stock discord zooming

### LINKED ISSUE #1018: [Feature Request] Zoom levels to match official app
### Discord Account

_No response_

### Motivation

I like everything in an app's UI to be just so, especially if I use the app for hours every day. Discord's official zoom levels are satisfying to me in terms of the list of servers fitting neatly in the available space, such that there are no partial server icons. Discord does this by defining zoom levels. Vesktop doesn't have the zoom levels defined in the settings UI or internally. So when you zoom in or out, it appears to go by 10% every time. This makes it impossible to reach zoom levels like 125% or 75%.
Screenshot_20241221_122120
Screenshot_20241221_122139

### Solution

I would like Vesktop's zoom levels to match the official app so that my server list looks nice.

### Alternatives

I can just have it at 120% instead of 125%, but in my opinion it doesn't look satisfying. Another option for the app would be to do 5% instead of 10% increments.

### Additional context

_No response_

### Request Agreement

- [X] This is not a plugin request
- [X] This is not a Vencord feature request

rushiiMachine:
Connects electron keybinds to Discord's internal zoom handling, and re-enables the zoom slider in appearance settings

FIXME:
- figure out what to do with incompatible default electron `viewMenu`

Fixes #1018

### DIFF
--- a/src/main/ipc.ts
+++ b/src/main/ipc.ts
@@ -93,6 +93,10 @@ handle(IpcEvents.MAXIMIZE, e => {
     }
 });
 
+handle(IpcEvents.SET_ZOOM_FACTOR, (e, factor: number) => {
+    e.sender.setZoomFactor(factor);
+});
+
 handleSync(IpcEvents.SPELLCHECK_GET_AVAILABLE_LANGUAGES, e => {
     e.returnValue = session.defaultSession.availableSpellCheckerLanguages;
 });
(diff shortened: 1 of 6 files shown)


WarpedEdge:
Is this getting merged yet? This is a must have for dual screens with different resolutions due to how Wayland sees multiple monitors as one :(

Kaydax:
This is similar to my PR #370, but this might be way better than my code

WarpedEdge:
Yeah its been waiting for a while now, and getting this merged in would be great.

lajczi:
@rushiiMachine can you fix merge conflicts?

PR #1134, opened 2025-03-19
feat: prevent mainWin from being used after destroyed

### LINKED ISSUE #808: [Bug] Unexpected JS error on quit after switching accs
### Discord Account

lumap

### Operating System

macOS Sequioa

_No response_

### Package Type

why is this not optional? vesktop is just a regular app on mac

The following window appears when `cmd+q`'ing (also known as completely closing) Vesktop:
Screenshot_2024-08-10_at_01 06 41

No errors appear

1. Have 2+ accs in your Vesktop
2. Switch from one account to another to another
3. Switch back
4. Quit the app with `cmd+q`

### Debug Logs

N/A, see image above

### Request Agreement

rushiiMachine:
Fixes #808

> The code triggering this seems to have changed a bit since this issue was created, but it seems to partially be caused by arrpc outliving the main browser window?
https://github.com/Vencord/Vesktop/commit/c9be6181644a329ce6556c49186e56e2b7d5e8e2#diff-53486cd162d85eb4faf85f2b936105851979cd772b7be0b7bc48ae85efbbb9e9L22-R22
(arrpc server tries to use `mainWin` to send an IPC event after it's already been destroyed?)

### DIFF
--- a/src/main/appBadge.ts
+++ b/src/main/appBadge.ts
@@ -42,7 +42,7 @@ export function setBadgeCount(count: number) {
 
             // circular import shenanigans
             const { mainWin } = require("./mainWindow") as typeof import("./mainWindow");
-            mainWin.setOverlayIcon(index === null ? null : loadBadge(index), description);
+            mainWin?.setOverlayIcon(index === null ? null : loadBadge(index), description);
             break;
     }
 }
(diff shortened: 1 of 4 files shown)


Vendicated:
mmh I'm not really a fan of this. Having to null check everywhere is annoying and in at least half of these places, the main window being destroyed would indicate a logic error so they shouldn't fail silently

It would be better if only the locations with listeners that can outlive the window check for status, so in this case arRPC

Vendicated:
https://github.com/Vencord/Vesktop/commit/524f8ff2776fa70c721c960cad3ece28b84c5ef2

PR #1139, opened 2025-03-29
add master switch to disable auto gain control

atar4xis:
#161

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -31,7 +31,7 @@ process.env.VENCORD_USER_DATA_DIR = DATA_DIR;
 function init() {
     app.setAsDefaultProtocolClient("discord");
 
-    const { disableSmoothScroll, hardwareAcceleration } = Settings.store;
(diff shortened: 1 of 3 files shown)


ukonen:
Compiled this change and it doesn't seem to work on windows, apologies for the lack of input but I have 0 knowledge when it comes to electron's screwery.
Windows

pap-stavros:
WebRtcAllowInputVolumeAdjustment is unreliable on windows...
You can patch getUserMedia at runtime like in #832 to actually disable agc.

trigger337:
Is this getting a fix? I'll have to switch back to Discord, because this is killing me

atar4xis:
hey @pap-stavros i put this up as a quick fix in case others ran into the same issue i did, i dont use windows myself, so if anyone else wants to weigh in, please feel free

pap-stavros:
@atar4xis hey thanks, your contribution was very helpful in leading me to the right direction as i don't really know react myself.

PR #1158, opened 2025-04-26
adds StartupVMClass to Linux AppImage .desktop entry

contrarybaton60:
This allows proper icon support and pinning in certain desktops (in my case Cinnamon)

image


Vendicated:
This is already what the WM class is, so adding this doesn't really do anything

Are you using the flatpak? Try [this PR](https://github.com/flathub/dev.vencord.Vesktop/pull/33) instead:
```sh
flatpak install --user https://dl.flathub.org/build-repo/176750/dev.vencord.Vesktop.flatpakref
```

contrarybaton60:
I'm using the AppImage

Vendicated:
Using original Vesktop, without this change, open a terminal and run
```sh
xprop WM_CLASS
```
then click on the Vesktop window. What is the output of the command?

contrarybaton60:
`WM_CLASS(STRING) = "vesktop", "vesktop"`

Samueru-sama:
>Using original Vesktop, without this change, open a terminal and run

I want to add that with or without this change the window class is always going to be the same `vesktop`.

`StartupWMClass` is used by the DE to know the name of the window class of the respective app, in the case of Cinnamon, KDE and others it is used to allow pinning apps to the dock, what you said makes me think that this would change the name of the window class when that isn't the case.

KDE still lets you pin the app when this is not set, but results in a bug where the icon duplicates when you open the application because it has no idea if the window that just opened is the one that belongs to that `.desktop` file, see https://github.com/ghostty-org/ghostty/pull/4930

PR #1160, opened 2025-05-09
Add switch to properly Disable Auto Gain Control on all platforms.

papstavros:
This PR addresses #161 by implementing a proper switch to disable automatic gain control.

- PR #1139 attempted to solve this using WebRtcAllowInputVolumeAdjustment, but this approach is unreliable on Windows
- PR #832 used a working approach (patching getUserMedia) but was hardcoded without UI controls

So this implementation creates a proper switch that actually disables agc on runtime. the flag used by 1139 is not reliable and does not work on windows, not sure why, ask chromium or electron. I do exactly what 832 does but i add a toggle switch in vesktop's settings aswell (like 1139). I have no experience with electron so I am not entirely certain it works but it compiles with no issues and seems to work on my arch machine. Oh also, sorry and thanks to the author of 1139.

Note: This is a workaround due to the unreliable behavior of WebRtcAllowInputVolumeAdjustment on Windows in Electron. This fix works for now but, it's fragile and could break if Electron or Chromium internals change. Ideally, AGC should be disabled with proper media constraints once those become reliable across platforms.

### DIFF
--- a/src/preload/disableAutogain.ts
+++ b/src/preload/disableAutogain.ts
@@ -1,0 +1,86 @@
+/*
+ * Disable Auto Gain Control
+ * Based on https://github.com/joeywatts/disable-autogain-control-extension/blob/master/disableAutogain.js
+ * Copyright (c) Joey Watts
+ * SPDX-License-Identifier: MIT
+ */
+
+function setLegacyChromeConstraint(constraint: any, name: string, value: any) {
+    if (constraint.mandatory && name in constraint.mandatory) {
+        constraint.mandatory[name] = value;
+        return;
+    }
+    if (constraint.optional) {
+        const element = constraint.optional.find((opt: any) => name in opt);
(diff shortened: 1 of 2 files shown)


pap-stavros:
Thanks for the clarification, doing that now

Vendicated:
thanks a lot for your efforts! but this is a poor implementation for various reasons:

1. This code should really not be in preload. Instead, it should be in renderer
2. It would be much better to integrate with Discord's switch instead of making a new switch
3. This leaves in a random console.log that really has no relevance to Vesktop users
4. This tries to patch deprecated and firefox specific functions, both of which are not even present in electron

For that reason I'll be closing this PR and just do it myself in a bit

Vendicated:
https://github.com/Vencord/Vesktop/commit/b17fef93ba3873ed67074e89c3b100a6c72243a9

PR #1169, opened 2025-08-12
fix: re-add tray icon to macOS

Allvaa:
I just noticed that when I open Vesktop on macOS, it has no tray icon. This was removed in #68 due to its super big size on the menu bar.

This PR fixed it by resizing the icon first instead of passing the icon directly to the tray constructor

image

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -130,8 +130,6 @@ function initTray(win: BrowserWindow) {
     if (process.platform === "darwin") {
         const icon = nativeImage.createFromPath(ICON_PATH)
             .resize({ width: 16 });
-        icon.setTemplateImage(true);
-
         tray = new Tray(icon);
     } else {
         tray = new Tray(ICON_PATH);
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -44,6 +44,7 @@ import { createSplashWindow, updateSplashMessage } from "./splash";
 import { makeLinksOpenExternally } from "./utils/makeLinksOpenExternally";
 import { applyDeckKeyboardFix, askToApplySteamLayout, isDeckGameMode } from "./utils/steamOS";
 import { downloadVencordFiles, ensureVencordFiles } from "./utils/vencordLoader";
+import { nativeImage } from "electron/common";
 
 let isQuitting = false;
 let tray: Tray;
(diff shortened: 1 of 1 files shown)


Vendicated:
should just apply this fix to all platforms

anyway I think the tray was removed because apps usually don't have trays on mac?

claration:
Vesktop has these options under the native menu bar, so having it in the trey wouldn't really be a benefit, at least apart from novelty.

image

Allvaa:
> should just apply this fix to all platforms

oo, so should I just resize it for all platforms? doesn't it break? I can't test for other platforms right now 

> anyway I think the tray was removed because apps usually don't have trays on Mac?

not sure, but docker, cf warp, etc has tray icon. I kinda find it useful tho for shortcuts, so why not

Allvaa:
> Vesktop has these options under the native menu bar, so having it in the trey wouldn't really be a benefit, at least apart from novelty.
> 
> image

oh wow, didn't notice this

Vendicated:
it would probably be a good compromise to simply make the tray setting opt-in on mac. that way users who want it can still enable it

claration:
The way Discord does it is that they only show their own trey icon when you're in a voice chat, to indicate that you're either unmuted/muted/or deafened along with quick actions regarding this. At least to me this would serve as an actual purpose of having a trey icon, instead of common quick actions you would find in the menubar. Though, I'm not really imposed on having different options for users.

Allvaa:
> The way Discord does it is that they only show their own trey icon when you're in a voice chat, to indicate that you're either unmuted/muted/or deafened along with quick actions regarding this.

Just checked, there is already a PR for that apparently

Allvaa:
> it would probably be a good compromise to simply make the tray setting opt-in on mac. that way users who want it can still enable it

now it is opt-in

src/main/mainWindow.ts
```diff
@@ -126,7 +127,14 @@ function initTray(win: BrowserWindow) {
...
     ]);
 
-    tray = new Tray(ICON_PATH);
+    if (process.platform === "darwin") {
+        const icon = nativeImage.createFromPath(ICON_PATH)
+            .resize({ width: 16 });
+        tray = new Tray(icon);
+    } else {
+        tray = new Tray(ICON_PATH);
+    }
```
Vendicated, line 136:
```suggestion
    const icon = nativeImage.createFromPath(ICON_PATH).resize({ width: 16 });
    tray = new Tray(icon);
```

antoniorodr:
Will this be implemented? I like to have a tray icon to see if ive some notification, because Ive the dock hidden.

antoniorodr:
For they who can still be interested, I ended up using [badgeify](https://badgeify.app) instead.

PR #1179, opened 2025-09-29
Add splash/tray customisation & change default splash to Tux

Vendicated:
Adds an API & UI for customisable Splash and Tray images

https://github.com/user-attachments/assets/78c41f73-e560-426a-9b66-8a8c92af5605


Disty0:
I am putting the 512x512 nearest resized version of shiggy here for people that prefers her over tux like me. They can customize the splash back to her if they want to:

shiggy_resize

NLZ:
> I am putting the 512x512 nearest resized version of shiggy here for people that prefers her over tux like me. They can customize the splash back to her if they want to:

Thanks, Vestop is not the same without her

MarioKart7z:
> I am putting the 512x512 nearest resized version of shiggy here for people that prefers her over tux like me. They can customize the splash back to her if they want to:

Thank you. I got REALLY WORRIED when i updated today and saw Tux instead.

I think removing her removed a lot of Vesktop's personality. I don't think this was the right call...

PR #1191, opened 2025-10-22
fix: bump electron to prevent macOS 26 WindowServer lags

ohaiibuzzle:
There was a bug in electron that was causing excessive lag on fruit based computer operating system version 26: https://github.com/electron/electron/pull/48376. 

It was fixed in Electron 38.2.0: https://releases.electronjs.org/release/v38.2.0. 
Vesktop is still using 38.0.0 on latest releases however, so that's a bit suboptimal.

This PR merely bumps the minimum version of version of Electron to prevent that from happening.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
         "@types/react": "19.2.1",
         "@vencord/types": "^1.13.2",
         "dotenv": "^17.2.3",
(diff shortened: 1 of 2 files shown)


Vendicated:
duplicate

PR #1194, opened 2025-10-26
fix: prevent updates from crashing on fedora (or other rpm based linux distributions) when zypper is not installed

liujustin604:
This bug was fixed upstream in electron-updater already, so bump the version
https://github.com/electron-userland/electron-builder/issues/9099


Vendicated:
https://github.com/Vencord/Vesktop/blob/main/patches/electron-updater.patch

PR #1198, opened 2025-10-31
feat: native module for idle detection on wayland

### LINKED ISSUE #396: Do not auto idle if user input is active
**Describe the bug**

Discord web goes to auto idle when the web page is unfocused for some time. Discord desktop client goes to auto idle when the user input is inactive for some time regardless of whether the client window is focused. The latter is much more sensible than the former. While I see there are limitations for a web page, I think this should be considered a bug because Vesktop is not behaving like what one would expect a desktop client to do.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS/Distro: NixOS 23.11
 - Desktop Environment (linux only): GNOME 45
 - Version: 1.5.0

**Command line output**

**Additional context**

T1mbits:
Adds a native node module for native idle detection on Wayland linux. To be used by the [NativeIdle plugin](https://github.com/Vendicated/Vencord/pull/3767).

Fixes #396


Vendicated:
Thanks for taking the time to work on this!

A few points:
- Adding a third party native binary isn't something I'm comfortable doing. It would have to be first party (i.e. part of libvesktop)
- This doesn't depend on any Vesktop specific things and has a lot of patches which have the potential to break regularely, so it would be much better to have this in the Vencord codebase instead of the Vesktop codebase. That way we can push fixes to it immediately without having to publish an entire new Vesktop release

T1mbits:
> Adding a third party native binary isn't something I'm comfortable doing

I figured that would be the case. I'll see if I can put the source code into a `/packages` dir now that I know how napi-rs works better than before. Again, I don't really know c++ so I'm not sure I'd be able to integrate it into libvesktop

> have this in the Vencord codebase instead of the Vesktop codebase

I'll see if I can figure out how to port it to Vencord, but it might take a little bit to figure out the codebase. Thanks for the quick response!

Vendicated:
Well it'd be a bit difficult to move it to Vencord while also having the native module, so I'm not sure about that. Don't worry about that part for now

T1mbits:
Moved the package into `/packges`, seems to work fine? Again, should probably test on other platforms and compositors. If you don't trust the prebuilt binaries you can feel free to rebuild them yourself.

nixRidge:
This works an absolute charm on Plasma Wayland! Thank you T1mbits.

nixRidge:
Actually, the only issue I've noticed is that Discord (Vesktop) doesn't auto-idle if my computer is on the lockscreen, which I've set Plasma to activate after five minutes. I don't know exactly where the problem is, but I'd guess that's something Plasma developers need to fix.

T1mbits:
it’s an unfixable problem for wayland, there isn’t a widely used protocol that allows clients to know if the system is locked or not so lockscreen idling is exclusive to windows and macos :/

nixRidge:
Got it, makes sense to me. So when the lockscreen is active (at least on Plasma as per the example), it just won't ever detect the user as idle?

In any case it's not a big issue.
I'm working around it by using the CustomIdle plugin and making it set my status to idle just before the screen locks (4,8 minutes), and in that case everything is fine. It even sets my status back to online when I move my mouse/click on the lockscreen.

T1mbits:
Not quite, it just doesn't know the lockscreen is active and as such will just continue acting as normal. That's why it sets the status back to online when you move your mouse on the lockscreen.

I've already moved most of the functionality to a Vencord plugin btw, this PR will pretty much just be for wayland module (I'll be pushing a commit later tonight that removes the functionality from this PR) so it'd be great if you could test out the plugin version instead :). The plugin already works with this vesktop patch though since it already exposes the wayland module.

nixRidge:
I'll do that tomorrow. Thanks for letting me know!

MatthewCash:
> Moved the package into `/packges`, seems to work fine? Again, should probably test on other platforms and compositors. If you don't trust the prebuilt binaries you can feel free to rebuild them yourself.

I don't see how this is better, its still just a prebuilt binary manually committed into the tree? I think the idea was that building libvesktop should also build the node addon and that no binaries are committed or downloaded.

T1mbits:
Not quite sure what you mean but the reason I didn’t just add onto libvesktop is because I don’t know how to program in C++ and I couldn’t get an environment working. I moved the module and source code into the main repo with the hopes that someone could rebuild and verify the code and binary directly themselves without forcing a third party npm dependency

MatthewCash:
> I moved the module and source code into the main repo with the hopes that someone could rebuild and verify the code and binary directly themselves without forcing a third party npm dependency

The problem is the prebuilt binaries though, Rust builds are generally not reproducible so verification would be difficult. I doubt the maintainers would be fine with that anyway, your changes would need to be integrated into the build system.

T1mbits:
The maintainers are also free to completely replace the binaries with their own compiled versions if they want. I wrote it in rust because, as I said, I couldn't write in C++ to extend libvesktop. If you want you can rewrite the module into libvesktop yourself to circumvent whatever issue you see with the way it currently is, but I unfortunately can't as of right now.

MatthewCash:
> The maintainers are also free to completely replace the binaries with their own compiled versions if they want. I wrote it in rust because, as I said, I couldn't write in C++ to extend libvesktop. If you want you can rewrite the module into libvesktop yourself to circumvent whatever issue you see with the way it currently is, but I unfortunately can't as of right now.

I might give it a shot if I get some time today :)

MatthewCash:
I added a prototype C++ port of your code to libvesktop in be04d6bef0ceb732d6c918cfbaf6f9138b818c96, so far completely untested or integrated, it builds though. It's missing the destructor for `IdleNotifier`, I'm not sure what the best approach for that is yet.

MatthewCash:
Fixed some bugs and now the latest commit on https://github.com/MatthewCash/Vesktop/tree/native-idle works for me.

eth-p:
> it’s an unfixable problem for wayland, there isn’t a widely used protocol that allows clients to know if the system is locked

If you don't mind adding logic for specific desktop environments, it is possible to check the lock screen state using dbus.

For KDE Plasma with kscreenlocker, you can query the lock screen state using dbus:
https://invent.kde.org/plasma/kscreenlocker/-/blob/Plasma/6.5/dbus/org.freedesktop.ScreenSaver.xml?ref_type=heads

```bash
dbus-send --dest=org.freedesktop.ScreenSaver --type=method_call --print-reply --session '/org/freedesktop/ScreenSaver' 'org.freedesktop.ScreenSaver.GetActive'
```
For GNOME, there appears to be an equivalent (found mentioned [here](https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/issues/787)):

```bash
dbus-send --print-reply=literal --session --type=method_call --dest=org.gnome.ScreenSaver /org/gnome/ScreenSaver org.gnome.ScreenSaver.GetActive
```

T1mbits:
Yeah I was aware of those when I originally wrote the module but I didn’t go through with it because I felt it was kind of a bandaid solution (and I also don’t use either environment so I couldn’t even test it lol). Might be worth asking @MatthewCash about it though since they seem to be capable of actually implementing something into libvesktop.

I’ll probably get around to looking at their implementation when I get back and might close or replace this pr with their rewrite too. Not quite sure yet though because there’s still some other stuff I never got around to fixing or testing but I appreciate the suggestion

MatthewCash:
I don't think Gnome even supports `ext_idle_notifier_v1` so if we're adding bandaid/nonportable solutions then calling Mutter's `GetIdletime` method might also be worthwhile.

T1mbits:
@MatthewCash not sure if you're still interested in this, but I've started working on this again and I've pulled your commit into my fork, hopefully that's okay with you. I haven't fully tested/recompiled your implementation yet as I've been busy reconstructing the powerMonitor events and basically remaking the plugin from scratch but I thought I should probably keep you updated on this.

MatthewCash:
I don't really have time to work on this anymore so I'm just hoping you can get this fixed :)

T1mbits:
Today I learned that I am not built for c++. I am a spoiled child and I need my cargo and my compiler and my docs.rs. 4 hours of headaches and I still can't get a destructor working for IdleNotifier. I'm probably just going to put back my rust implementation with proper integration into the buildsystem (with docker and build scripts for prebuilts and the likes), unless somebody else who is actually capable and has free time can patch things up.

I'll wait till the end of the week before doing anything but if I do I'll probably try to implement the protocols @eth-p mentioned earlier.

mati865:
I suppose you have seen https://github.com/MuffinTastic/wayafknext-monitor?
It appears to handle both Wayland and Mutter sides of things.

I can try to find some time to help with Rust implementation.

T1mbits:
I'm going to be completely honest, I've never seen that before. I was talking about the prior package I made ([you can see it here](https://github.com/T1mbits/Vesktop/tree/d66a7c96d85d1fdc4a85a342c86af0b860d24b92/packages/wayland-protocols)) which Matthew replaced with a direct libvesktop addition due to concerns about reproducibility and lack of proper integration. I'm pretty sure I can set up something like another dockerfile build script to pin rustc/cargo properly for reproducibility on prebuilt builds, so hopefully that's less of a concern?

Looking at wayafknext, it takes a very different approach as I made a native node module to work inside of vesktop whereas wayafknext-monitor seems to be its own dedicated binary which talks over a dedicated ipc socket since it's an external plugin.

If you do want to help extend my module for plasma and mutter before I properly integrate it, the original source code is over at https://github.com/T1mbits/vesktop-wayland-protocols. I did change it a bit when I put it into the vesktop repo I think, but I don't really remember.

mati865:
Rust binaries are usually reproducible OOTB, although there are some edge cases. I don't have experience with napi, but chances are it might be reproducible already. Making it easier to build for non-Rust people might be sensible though.

Given that Gnome's Mutter is probably the only popular compositor missing `ext_idle_notifier_v1`, it sounds like we should probably handle it when already considering zbus for KDE's lockscreen.
I won't be able to look into it before the next week.

I might as well CC @MuffinTastic here, as they might want to contribute parts of their implementation here.

MuffinTastic:
I'm wondering how relevant all of this is now, considering Discord's latest announcement video regarding Linux:

https://youtu.be/BwNfmazmU4o?t=184

> We also fixed a bunch of smaller issues. [...] Wayland idle protocol support, which means you can step away from your desktop, and still get notifications to your phone. [...]

And indeed, I've heard from some friends who don't use Vencord/Vesktop (and by extension don't use my plugin) that they've noticed their idle detection has started working in vanilla Discord on Linux.

I want to hear a few more reports, especially from the GNOME side of things since all my friends use KDE, but I expect to be able to officially retire the native idle monitor side of my plugin.

T1mbits:
Well that's great news for people who use the official client and are willing to switch to it, but I personally don't use the official client and probably won't as long as Vesktop's an option for me, regardless of how good its linux support becomes in the future. Idle detection only works on the official client itself regardless, so Vesktop is completely unaffected by this anyways.

I just wanted feature parity with the official client, and if this is a recent update that means this PR was ahead of its time lol. Either way, while it will probably take me a comically long time since I keep getting distracted, I don't plan on stopping.

Just a note, (not sure if anybody cares but maybe @mati865 does?) I'm going to probably force-push/reset this branch when I get the original package back in here since the history is just a giant mess, but I am currently binge watching and procrastinating on a few responsibilities so that might be a little while.

MuffinTastic:
> Idle detection only works on the official client itself regardless, so Vesktop is completely unaffected by this anyways.

Ah yeah, true. Disregard my previous comment...

> 4 hours of headaches and I still can't get a destructor working for IdleNotifier.

I wasn't familiar with Napi before reading through the drafts here, but Napi's ObjectWrap API immediately caught my eye. It seems as though the lifetime for objects created through that API are fully controlled by V8, so for the destructor to run (through the JS finalizer) you'd need the IdleNotifier object to get garbage collected somehow. Which would be fine by itself, but it also seems as though it's not guaranteed that V8 garbage collects on process exit. So I think you end up with the IdleNotifier object lingering until the bitter end.

If it were me, to get around all this I'd add explicit JS-accessible start()/stop() functions to IdleNotifier, and remember to call stop() before process shutdown, maybe through `process.on("exit", ...)`. JS isn't my forte though...

T1mbits:
I’m completely out of my element with C and C++, so unless someone is willing to fix and extend libvesktop to encompass everything everyone has spoken about, I’d still prefer to put the rust module back in place since it’s something I can at least work on myself. Regarding destructors and the IdleNotifier objects, I’m pretty sure on wayland it’s never unloaded once loaded by the plugin, but only one can be loaded per Vesktop instance. I also don’t see a point in a dedicated stop/start system since that requires IPC, which is a bit excessive. If you restart Vesktop entirely with the plugin disabled, IdleNotifier is not ever created, so no penalty for anyone who doesn’t want it. The reason a destructor was so important is because when Vesktop tried to close it would cause an exception/undefined behaviour with the hanging thread and wayland connections, which doesn’t happen on the rust implementation. I’m realizing I’m really glazing rust hard, but it just boils down to the fact that nobody has offered to fix the libvesktop version and C just confuses the heck out of me

mati865:
> Just a note, (not sure if anybody cares but maybe @mati865 does?) I'm going to probably force-push/reset this branch

No worries.

> Well that's great news for people who use the official client and are willing to switch to it

Yeah, the ship has sailed for me as well. Moved fully to Vesktop last week and removed official app 3 days ago.

mati865:
Perhaps I should give a status update. I'm struggling to reliably detect when the screen is locked in my DE that is COSMIC.

`org.freedesktop.ScreenSaver.GetActive` API is a nonstandard extension implemented at least by KDE: https://github.com/pop-os/cosmic-idle/issues/5
Based on a quick query through the code it's not implemented by Sway/wlroots, Niri, Jay and Hyprland.

I've also tried to listen on `org.freedesktop.login1.Session` interface but on COSMIC only lock events appear there: https://github.com/pop-os/cosmic-greeter/issues/266

I'm not sure if there is a point in holding this back - even without session lock handling this is an improvement and further improvements can happen later on.

T1mbits:
Oh I completely forgot I had planned to work on this over the weekend T-T. Maybe I do have commitment issues. Anyways, as far as I’m aware it’s by design that lockscreens can’t be detected on wayland so I’m still not sure if it’s something worth pursuing. If you do get it to work that’d be great though! I’m also curious how you’re implementing the lockscreen detection into the powerMonitor api if you’ve done anything on that front yet. I want to slightly tweak the current isIdle api but I keep forgetting and losing all my free time…

Issue #629, opened 2024-05-28
[Bug] Very low FPS when Screensharing


PacketAuditor:
### Operating System

EndeavourOS (Wayland)

### Linux Only ~ Desktop Environment

KDE

### What happens when the bug or crash occurs?

When screensharing, the quality is not ideal, but the biggest issue is the framerate is horrible. ~2-10 FPS.

### What is the expected behaviour?

The frame rate selected is respected.

### How do you recreate this bug or crash?

1. Start Screensharing

### Debug Logs

```
[p@p-pc ~]$ vesktop
[arRPC > ipc] listening at /run/user/1000/discord-ipc-0
[arRPC > websocket] listening on 6463
[arRPC > process] started
src/gbm_drv_common.c:131: GBM-DRV error (get_bytes_per_component): Unknown or not supported format: 808530000

src/gbm_drv_common.c:131: GBM-DRV error (get_bytes_per_component): Unknown or not supported format: 808530000

src/gbm_drv_common.c:131: GBM-DRV error (get_bytes_per_component): Unknown or not supported format: 808530000

src/gbm_drv_common.c:131: GBM-DRV error (get_bytes_per_component): Unknown or not supported format: 808530000

src/gbm_drv_common.c:131: GBM-DRV error (get_bytes_per_component): Unknown or not supported format: 808530000

'loop->recurse > 0' failed at ../pipewire/src/pipewire/thread-loop.c:426 pw_thread_loop_wait()
[2024-05-28 17:38:33.624] [venmic] [info] [patchbay] (get) running venmic 3.4.2
[2024-05-28 17:38:33.629] [venmic] [info] [patchbay] (handle) found default metadata: 37
[2024-05-28 17:38:33.629] [venmic] [info] [patchbay] (meta_update) speaker name: "alsa_output.pci-0000_0b_00.4.iec958-stereo.2"


```

### Request Agreement

- [X] I have searched the existing issues and found no similar issue
- [X] I am using the latest Vesktop and Vencord versions
- [X] This issue occurs on an official release (not just the AUR or Nix packages)

PolisanTheEasyNick:
Can you please send screenshot of stream debug? You can get it by:
1. Enabling Developer mode in Discord settings
2. Starting Stream
3. Clicking on "Voice connected" green text and "Debug" upper screenshare button in bottom left corner
4. Picking "Output" under "Debug RTC: Stream"

Also, please specify which content type you are using (smoothness or clarity).

[...]

Issue #525, opened 2024-04-20
Very low FPS when streaming (Windows 11)


deko0:
**Describe the bug**

Trying to stream anything results in having a framerate of about one, maybe less. Switching to a browser, regular discord, or discord with Vencord, the framerate and streams work as expected. Framerate does not fluctuate, it is constantly that.

**To Reproduce**

  1) Try streaming anything, in any resolution/fps combo
_2) My stream is washed 💔_

**Expected behavior**

Stream framerate to be what i choose it to be

**Desktop (please complete the following information):**

Windows 11 23H2

**Command line output**

doesn't do anything on windows, i assume this is for Linux?

**Additional context**

My gpu drivers are updated. Tried reinstalling using DDU. It's not either my connection or my pc. 
Just in case this is needed, here are my specs:
GTX 1060 low profile 3gb
Intel core i5 8400
16GB DDR4-2666mhz ram
Vesktop and Discord installed on an m.2 nvme SSD

PolisanTheEasyNick:
Most probably you are installed Vesktop from Vesktop-Setup-1.5.1.exe installer from Releases, which is not actual upstream version of Vesktop. There was a bug when framerate is 30 fps; resolution picking from display resolution by standard; and bitrate capped to 2500 kbit/s. All this was fixed in #489.
Can you re-produce issue using [building from source](https://github.com/Vencord/Vesktop?tab=readme-ov-file#building-from-source)?

deko0:
tutorial is for Linux. no idea how to those things in windows though

> Most probably you are installed Vesktop from Vesktop-Setup-1.5.1.exe installer from Releases, which is not actual upstream version of Vesktop. There was a bug when framerate is 30 fps; resolution picking from display resolution by standard; and bitrate capped to 2500 kbit/s. All this was fixed in #489. Can you re-produce issue using [building from source](https://github.com/Vencord/Vesktop?tab=readme-ov-file#building-from-source)?

How do i do that stuff on Windows? I cloned the repo but I don't really know how to build it.

PolisanTheEasyNick:
> How do i do that stuff on Windows? I cloned the repo but I don't really know how to build it. and yeah, i did in fact install 1.5.1

You need to install pnpm firstly: https://pnpm.io/installation

deko0:
> > How do i do that stuff on Windows? I cloned the repo but I don't really know how to build it. and yeah, i did in fact install 1.5.1
> 
[...]

PR #1202, opened 2025-11-15
fix: screenshare performance with adaptive constraints

BenHodgsonR2R:
Issues:
https://github.com/Vencord/Vesktop/issues/629
https://github.com/Vencord/Vesktop/issues/525

### Problem
When sharing games or other applications via screenshare, users experience significant performance degradation in the shared application and/or the stream itself. This is particularly noticeable when using the "Prefer Clarity" (detail) content mode.

The issue stems from the use of `min` constraints and the `advanced` constraint array. According to the [W3C Media Capture and Streams specification](https://w3c.github.io/mediacapture-main/#constrainable-interface):

> **Constraints which are specified using any or all of `max`, `min`, or `exact` are always treated as mandatory.** If any constraint which uses one or more of those can't be met when calling `applyConstraints()`, the promise will be rejected.

Additionally, bare values in the `advanced` array are treated as `exact` constraints, which creates rigid requirements that the capture system must work hard to maintain.

When the system cannot maintain mandatory minimums, it will:
- Duplicate frames to hit minimum frame rate requirements
- Block/wait for frames from the capture source
- Apply frame rate conversion with CPU overhead
- Put additional load on the shared application's GPU/CPU

### Solution
This PR introduces **content-aware adaptive constraints** that match the user's selected content hint:

**"motion" mode (Prefer Smoothness):**
- Uses `min` constraint for frame rate to ensure smooth playback
- Uses `ideal` constraint for resolution to allow adaptive scaling

**"detail" mode (Prefer Clarity):**
- Uses `ideal` constraints for both frame rate and resolution
- Allows the capture system to gracefully adapt when the shared application is under load
- No mandatory minimums that force frame duplication or blocking

**Both modes:**
- Removed the `advanced` constraint array that was forcing exact resolution matches
- Lets basic constraints handle adaptation naturally per W3C spec guidance

With `ideal`-only constraints, the browser targets the specified values but can drop below them without penalty, removing the performance overhead on the shared application.

### Technical Details
The key insight is that `ideal` values are treated as optimization targets, not mandatory requirements. From the spec:

> **ideal** - A decimal number specifying an ideal value for the property. If possible, this value will be used, but if it's not possible, the user agent will use the closest possible match.

This allows the capture system to balance quality vs. performance dynamically based on what the shared application can actually provide.

### Testing
Tested with game streaming on Linux. Performance improvements are noticeable in "detail" mode with no regression in "motion" mode behavior.

### DIFF
--- a/src/renderer/patches/screenShareFixes.ts
+++ b/src/renderer/patches/screenShareFixes.ts
@@ -39,7 +39,8 @@ if (isLinux) {
         const constraints = {
             ...track.getConstraints(),
             ...getContentResolutionSettings(contentHint, { width, height }),
-            frameRate: getContentFrameSettings(contentHint, frameRate)
+            frameRate: getContentFrameSettings(contentHint, frameRate),
+            resizeMode: "none"
         };
 
         track


Vendicated:
This doesn't really seem right. The only thing that is definitely right is not setting a minimum frame rate for the detail preset

All the other changes seem like they would cause more issues than harm. With your changes, detail will set the minimum for the 1080p preset to 1920x1080. What if you share a window that's smaller? Or you have a 16x10 monitor, etc.

PR #1209, opened 2025-11-29
feat: add taskbar flashing on new notifications

futamebore:
Added taskbar flashing when a new notification arrives.

This is a feature that exists in the official client and is useful for people with OLED setups that use the common burn-in prevention methods (auto-hiding the taskbar and having a short screen timeout). With this change, you can easily see that someone messaged you after waking the screen up, or if you simply weren't looking at the screen when the notification arrived.


Vendicated:
Well there's one small issue: The app immediately flashes if you start it with existing notifications

Vendicated:
Well Discord just calls flashFrame when a Notification is shown. Also they only do it on Windows for some reason.

image

image

PR #1213, opened 2025-12-11
feat: add voice settings management in arRPC

kymotsujason:
- Added IPC commands for setting and getting voice settings and selected voice channel.
- Introduced voice state updates and dispatching mechanisms in the server and worker.
- Updated pnpm-lock.yaml with new patch hash for arrpc.

The intention here was to get StreamController (Linux Stream deck software) working with Vesktop. Specifically, using the Discord plugin and the mute/deafen buttons. 

The initial mute startup state doesn't get communicated, but you can just join a call or unmute to fix.

### DIFF
--- a/patches/arrpc@3.5.0.patch
+++ b/patches/arrpc@3.5.0.patch
@@ -1,5 +1,5 @@
 diff --git a/src/process/index.js b/src/process/index.js
-index 389b0845256a34b4536d6da99edb00d17f13a6b4..f17a0ac687e9110ebfd33cb91fd2f6250d318643 100644
+index f445024f14ecfe5ee0eb70fe42535591c76838dc..4f089d035202a611ad5f18a74925e8d273a095b4 100644
 --- a/src/process/index.js
 +++ b/src/process/index.js
 @@ -5,8 +5,20 @@ import fs from 'node:fs';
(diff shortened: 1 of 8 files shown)


Vendicated:
Hi thanks for working on this!

But don't patch arrpc please. We have a fork that we were planning to maintain: https://github.com/Vencord/arrpc

You can just open a pull request there :) We will eventually publish it at @vencord/arrpc, (or we could just vendor it into this repo, I honestly don't really care). For now you can just use a local install (`pnpm install ../arrpc`) to test with it

Also long term we plan to stop using arRPC and instead reuse Discord's own handlers, so in case you want to add even more features, it'd be best to first come discuss in our Discord before wasting your time on work that might be obsolete (for now, this PR is fine)

kymotsujason:
Hi, thanks for taking a look.

I didn't realise you guys had a fork of it, it's definitely a better idea to just send the arRPC patch changes over there instead. I'll remove the arRPC patch from this PR and put this in draft for now. I also realised, is there no authorisation window in the code? I was testing a clean start and didn't see one.

I think using Discord's own handlers is probably the best way forward for sure, I just needed mute and deafen features for my specific use case. Figured it might help the 10 other people who might also have similar hardware haha.

Vendicated:
What do you mean by authorisation window?

Discord's login form? That's handled by the discord website

kymotsujason:
Oh wait nvm, that does work, I just tested with Discord activity and the auth thing pops up. I'm probably just missing something here.

Vendicated:
Did you get stuck on anything? Or just been preoccupied with other things? You can come join our Discord and hop in the vesktop-development channel (open a modmail ticket to get access) if you want to discuss further

kymotsujason:
I'm not really satisfied with the reliability of this method. When it works, it's great. However, the connection is flaky. Restart discord and stream controller and it might break, do it again and it might work. I'll probably just close the PR by next week if I don't make more progress. 

Your roadmap plans are better. I don't think there are enough users who'd benefit from this addition to merge it haha.

Issue #1210, opened 2025-12-02
vesktop is not selecting the correct scene in obs


nyxar77:
### Content

Vesktop: v1.6.1
OS:  - system: `"x86_64-linux"`
 - host os: `Linux 6.12.59, NixOS, 25.11 (Xantusia), 25.11.20251130.8bb5646`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.31.2`
 - channels(root): `"nixos-25.05, unstable"`
 - nixpkgs: `/nix/store/dgp9wmspgm6zhl66f0c81rvg3v7wwbsb-source`

when i choose a scene in obs and select the option "start virtual camera" vesktop fails to detect the same scene so it defaults to using the laptop's camera, this wasn't an issue on discord so it's on vesktop only

obs:

<img width="1916" height="993" alt="Image" src="https://github.com/user-attachments/assets/6f51c421-e007-4b1b-9dd5-ad4c7cee6ae1" />

vesktop scene:

<img width="892" height="762" alt="Image" src="https://github.com/user-attachments/assets/89b5e02a-fcb7-4ee3-a717-4a82bcc4d388" />

same scene on discord:

<img width="1063" height="812" alt="Image" src="https://github.com/user-attachments/assets/b2877fde-3f5b-4efc-876c-d39ecee44e80" />

works as expected

MathisP75:
I have the same issue when trying to select another camera on my laptop. Even if I choose the USB camera, it uses my laptop's camera.

MathisP75:
I confirm that I can reproduce the issue with the OBS virtual camera too.

nyxar77:
nixos finally received the latest updates on the unstable channel, and i wanna tell you that the changes @Creationsss made did not fix the issue, at least for me

<img width="1450" height="977" alt="Image" src="https://github.com/user-attachments/assets/7de3ec30-a740-4214-b9c2-887bc59900f0" />

and i can confirm this issue is specific to vesktop and not discord in general

PR #1215, opened 2025-12-18
fix: enforce exact deviceId constraint for video devices

Creationsss:
fixes video device selection not working for virtual cameras e.g. OBS Virtual Camera.

this just converts string deviceId to exact constraint format to prevent Chromium from silently falling back to the default camera

should fix https://github.com/Vencord/Vesktop/issues/1210


Forceres:
any updates?

PR #1223, opened 2026-01-27
fix: disable WaylandWpColorManagerV1

Mylloon:
This fixes this color issue that occurs on Wayland:

https://github.com/user-attachments/assets/a8908888-d9a6-440b-a79b-3a5a8aa82bee

Linked issue: https://github.com/electron/electron/issues/49566

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -28,6 +28,8 @@ process.env.VENCORD_USER_DATA_DIR = DATA_DIR;
 
 const isLinux = process.platform === "linux";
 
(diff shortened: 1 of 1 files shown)


ssareta:
Should we gate it with an environment variable instead of flat disable? Colour management is a useful feature, could be good to let users enable once it gets fixed upstream + for wayland compositors that implement it properly (e.g. HDR image/video playback, as a random example)

Mylloon:
Upstream fixed it!

Vendicated:
> Upstream fixed it!

No they didn't actually lmao, it's still an issue on latest Electron (43.0.0), but not an issue on the version Vesktop uses

PR #1224, opened 2026-01-28
Implemented open.asar mods

morawskidotmy:
Added open.asar implementation with a toggle option in vesktop settings, ported:
-Performance/Battery mode
-Thier tracking filter
-Typing indication disable
-DOM optimization


Vendicated:
no

morawskidotmy:
👍

Vendicated:
I appreciate your contribution but all of these additions are either redundant or problematic:

> tracking filter, Typing indication disable

Both of these already exist as Vencord plugins

> DOM optimization

This is complete voodoo

> Performance/Battery mode

Chromium performance flags are very system dependant. For some users they may improve things, but for other people it might worsen performance or even cause nasty issues. As such, providing presets is pretty pointless. There are various websites that document useful flags (e.g. arch wiki), the vesktop wiki could link to those

> Custom user flags

This likely has security issues. There are thousands of chromium flags, and letting the browser control them is likely problematic and might allow a sandbox escape

morawskidotmy:
thx for replying, i was just commiting coz ill use that anyway and i dont want to add too much stuff like i did to other forks. totally understendable :> have a nice day or night :>

PR #1239, opened 2026-04-05
Vencord application media file optimization

thegreatGreenstar:
This optimizes SOME of the files here, overall with lossless optimization. 
The exception is to build/icon.svg and the webp
The webp has had timing of it slightly adjusted, though the frames themself have no lossy changes otherwise.

The png use https://github.com/oxipng/oxipng via https://ezgif.com/optipng and the other content is optimized through similar means
The svg though is instead optimized via https://svgomg.net


Vendicated:
what's the point of these changes

thegreatGreenstar:
> what's the point of these changes

Filesize reduction

Vendicated:
this is a desktop app with local files, not a website. there is no point hyperoptimising for 15kb savings via lossy compression...

thegreatGreenstar:
I'll revert the optimization on the svg and webp so it's all lossless changes.

PR #1242, opened 2026-04-12
Fix Linux screenshare state drift and sender recovery during viewer attach/rejoin

alper-han:
## Summary
- keep the selected Linux screenshare profile aligned across capture constraints, `videoStreamParameters`, and RTP sender parameters
- resync the active screenshare path after stream start, profile changes, and `replaceTrack()`
- add a guarded recovery path for viewer attach/rejoin when the outbound sender still appears stalled after the normal resync path

## Context
This started from Linux NVIDIA VAAPI/NVENC validation, especially `Prefer Clarity` sessions.

Before this patch, that setup could not reliably recover viewer attach/rejoin. The sender could remain alive while the effective screenshare state drifted across capture constraints, connection stream parameters, and RTP sender configuration.

Related driver-side validation context:
- https://github.com/elFarto/nvidia-vaapi-driver/pull/425

## Notes
The failure was first reproduced during NVIDIA VAAPI validation, but the fix belongs in Vesktop's Linux screenshare path rather than in a vendor-specific branch.

The recovery path is intentionally conservative. It first reapplies the selected profile and resynchronizes sender state, then only escalates to `replaceTrack()`-based recovery if the sender still appears unhealthy after that safer resync step.

This patch only changes Linux behavior.

### DIFF
--- a/src/renderer/components/ScreenSharePicker.tsx
+++ b/src/renderer/components/ScreenSharePicker.tsx
@@ -42,6 +42,14 @@ import { SimpleErrorBoundary } from "./SimpleErrorBoundary";
 
 const StreamResolutions = ["480", "720", "1080", "1440", "2160"] as const;
 const StreamFps = ["15", "30", "60"] as const;
+export const screenShareBitrateMin = 500000;
+export const screenShareBitrateMaxFloor = 8000000;
+export const screenShareBitrateTarget = 600000;
+const screenSharePixelCountGuardKey = "__vesktopScreenSharePixelCountGuard";
+const screenShareVideoStreamParametersGuardKey = "__vesktopScreenShareVideoStreamParametersGuard";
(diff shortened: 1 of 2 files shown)


Vendicated:
Thanks for your contribution

Please explain better what this pull request is trying to solve. Is this solving a bug? Why are so many changes (thousands of lines) necessary?

VMFortress:
I only took a brief look at the changes but I was curious to test it. 

Whether intentional or not, these changes seem to resolve the issue with screenshare bitrate plummeting to near zero when audio is included which has stopped me from using Vesktop for years now.

Not perfectly fixed as it will still sometimes it still drop lower and drop frames, especially at higher resolutions/refresh rates, but it is a vast improvement. I'll have to look through these changes more thoroughly.

Issue #792, opened 2024-08-02
[Bug] Vesktop disconnects while i'm streaming


Steccah:
### Operating System

Arch Linux

### Linux Only ~ Desktop Environment

KDE on Wayland

### Package Type

AppImage and aur

### What happens when the bug or crash occurs?

When i stream something in whichever quality, after a while, it disconnects. It doesn't disconnects by discord, i just can't hear others, can't talk and others see my stream freezed. If i exit the call and re-enter i can use discord just fine, if i stream again i will be disconnected faster, and every time discord disconnects me i will be disconnected faster than the previous time.

### What is the expected behaviour?

I expect not to be disconnected

### How do you recreate this bug or crash?

1. stream something
2. wait ~10 minutes
3. you can't hear others and others can't hear you

### Debug Logs

```
❯ ./Vesktop-1.5.3.AppImage
Checking for update
[arRPC > ipc] listening at /run/user/1000/discord-ipc-0
[arRPC > websocket] listening on 6463
[arRPC > process] started
Update for version 1.5.3 is not available (latest version: 1.5.3, downgrade is disallowed).
checkForUpdatesAndNotify called, downloadPromise is null
[138460:0802/141224.589744:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 1 times!
[138460:0802/141224.595884:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 2 times!
[138460:0802/141227.480599:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 3 times!
'loop->recurse > 0' failed at ../pipewire/src/pipewire/thread-loop.c:425 pw_thread_loop_wait()
[2024-08-02 14:14:45.223] [venmic] [info] [patchbay] (handle) found default metadata: 38
[2024-08-02 14:14:45.223] [venmic] [info] [patchbay] (meta_update) speaker name: "alsa_output.pci-0000_0e_00.4.analog-stereo"
[2024-08-02 14:14:45.223] [venmic] [info] [patchbay] (get) running venmic 6.1.0
```

### Request Agreement

- [X] I have searched the existing issues and found no similar issue
- [X] I am using the latest Vesktop and Vencord versions
- [X] This issue occurs on an official release (not just the AUR or Nix packages)

Steccah:
Refocusing does nothing

gibarel1:
[...]

Issue #320, opened 2024-01-03
Vesktop pauses when not focused


s-cd:
When you have vesktop unfocused on fedora sliverblue 39 after a few minutes the program seems to "hibernate" and disconnects from any calls until i refocus again and it reloads. After checking this doesn't seem to happen with the official app.

Vendicated:
please provide more info

how did you install vesktop
what desktop environment do you use
what vesktop version

s-cd:
Flatpak
Hyprland
Vencord 5dee2e8 (Vesktop V0.4.4)

Curve:
I have the same issue on Arch Linux with vesktop from the aur - I can't seem to reproduce this consistently though, however it happens fairly often

Vendicated:
this is really strange because i've never had this happen. likely related to power saving. we should check what flags Discord passes and see if any of them could relate to this issue

Madmaxispog:
im having the same issue

dowttie:
same issue here. i'm on arch, kde plasma, got vesktop from the aur.
it seems to happen inconsistently as described before.
worth mentioning that i started using vesktop last week and i had this with vencord on regular discord before as well.

Noden2048:
Same issue here
Everything goes quiet and after refocusing, the connection status says something along the lines of "Reconnecting...", logo then flashes and then it reconnects to the call 

x11 Debian, Vesktop (version 1.5.0 4bb0db5), installed the Vesktop app from the .deb

purepani:
I used to have this not happening, but after updating to the most recent plasma6 rc, ive had this issue. Just thought I'd give some more info in case it's helpful

s-cd:
okay i thought this problem solved itself but I realized that this only happens in calls, maybe try that?

adpkjapkbpkadnbpandb:
happens to me whenever i receive a notification, could be the issue, can someone try that?

QuietTR55:
I have the same issue, I Use Pop_os and I have enabled "Run in background" for this app but still, it still suspends when out of focus.

Curve:
> I have the same issue on Arch Linux (with KDE) with vesktop from the aur - I can't seem to reproduce this consistently though, however it happens fairly often and always when Vesktop is minimized or sometimes even when it's merely unfocused.
> 
> When regaining focus (or being restored) the discord (re-)loading screen is shown and shortly after I'm thrown back into discord as if it just started
> 
> I've been having this issue with any vesktop version as well

[...]

PR #1249, opened 2026-04-26
Auto-reload renderer when it crashes

1letme:
## Summary

When Vesktop's renderer process dies (commonly during long screen-share sessions, see #792), the main Electron process keeps running but the window appears frozen because there's no UI to update. The existing `render-process-gone` handler only logs the event:

```ts
mainWin.webContents.on("render-process-gone", (event, details) => console.log(details));
```

This PR makes that handler reload the renderer when the failure looks recoverable, so the user gets a brief interruption instead of a dead window that requires a full app restart.

## Diagnostic evidence

Captured from a session that froze after ~20 minutes of full-screen 4K streaming on NVIDIA + Wayland (Hyprland), running Vesktop with `--enable-logging=stderr --v=1`:

**The renderer crashed — it didn't just hang:**

```
[2791694:0426/025109.179228:VERBOSE1:.../media_stream_manager.cc:1589] AOC::StopStream => (error_during_callback=false)
Renderer process crashed - see https://www.electronjs.org/docs/tutorial/application-debugging for potential debugging information.
[2791694:0426/025109.194965:INFO:.../shared_screencast_stream.cc:253] PipeWire stream state: paused
[2791694:0426/025109.195025:INFO:.../shared_screencast_stream.cc:253] PipeWire stream state: unconnected
```

**`ps` after the freeze confirms the renderer process is gone** while the main Electron process is still alive in `do_sys_poll` — exactly the state that produces a "frozen window" from the user's POV.

**WebRTC was already starving for ~20 min before the crash** (3840×2160 input, only ~7 fps making it out, ~60% drop rate sustained):

```
VAdapt Drop Frame: scaled 0 / out 26905 / in 65515 ... Input: 3840x2160 ... Output fps: 7
```

So the failure mode is: capture/encode pipeline overload → renderer becomes unresponsive → Chromium kills it → main process has nothing to draw → window appears frozen until full app restart.

VRAM was stable (~5 GB / 6 GB free) the whole time — this is not OOM, it's the renderer-watchdog killing an unresponsive process.

## Behavior

- **Auto-reloads** on `crashed`, `oom`, `abnormal-exit`.
- **Doesn't reload** on `clean-exit`, `killed`, `launch-failed`, `integrity-failure` (where reload is wrong or unsafe).
- **Loop-breaker:** if a second crash happens within 30s of the previous one, gives up so a renderer that crashes on load can't reload forever.
- Logs both the crash details and the reload attempt at error/info level.

## Why this is worth shipping even without a root-cause fix for #792

The actual cause (Chromium's capture pipeline + NVIDIA Wayland encode being too slow to keep up at high resolutions) is upstream Electron / driver territory and the maintainer position on similar issues (#320) has been "wait for electron." That's reasonable.

But it leaves users with a UX where a 30+ minute stream silently turns into a dead window with no indication of what happened. This patch doesn't fix the cause — it just makes the failure mode recoverable: stream blips for a couple seconds, renderer comes back, user knows what happened. Tiny surface area, low risk.

## Test plan

- [ ] Stream a screen long enough (or under enough load) to trigger a renderer crash; confirm the window auto-reloads instead of freezing.
- [ ] Force a renderer crash via DevTools (`process.crash()` in the renderer) and confirm reload.
- [ ] Verify normal app close still works (no spurious reload on `clean-exit`).
- [ ] Verify CI build passes.

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -38,6 +38,7 @@ import { downloadVencordFiles, ensureVencordFiles, vencordSupportsSandboxing } f
 import { VENCORD_FILES_DIR } from "./vencordFilesDir";
 
 let isQuitting = false;
+let lastRendererCrashAt = 0;
 
 applyDeckKeyboardFix();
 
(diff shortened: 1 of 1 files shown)


Vendicated:
don't send ai generated pull requests

PR #1250, opened 2026-04-26
feat: protect Discord tokens using OS-level encryption (safeStorage)

Polyfish0:
This PR introduces token protection to mitigate common token grabbers that extract credentials directly from disk (e.g. LevelDB or Cookies SQLite).

### Overview

When Electron's `safeStorage` is available, Discord tokens are no longer stored in plaintext. Instead:

* localStorage (`token`)

  * Intercepted via a proxy
  * Stored as an encrypted blob under a separate key
  * Decrypted transparently at runtime

* Cookies (`tokens`)

  * Intercepted via a `document.cookie` proxy
  * Encrypted before being persisted by Chromium
  * Decrypted on read

The plaintext token never touches disk.

---

### Behavior & Design Decisions

* Fail closed:
  If encryption fails, plaintext storage is refused. This may result in a logout, which is considered preferable to leaking tokens.

* Session fallback (no safeStorage):
  If encryption is unavailable, the `tokens` cookie is downgraded to a session cookie (removes `Expires` / `Max-Age`) to avoid persistence on disk.

* Migration:
  Existing plaintext `tokens` cookies are migrated to encrypted storage on startup.

* In-memory cache:
  Tokens are cached in memory to maintain compatibility with Discord’s expectations.

* Hardening:

  * `document.cookie` proxy is defined with `configurable: false` to prevent redefinition
  * Only plaintext values are migrated (avoids double encryption)
  * Decryption failures fall back to raw values, triggering re-auth instead of exposing errors

---

### Scope

This protects against:

* Token grabbers reading LevelDB or Cookies DB
* Malware scanning disk for plaintext tokens

This does not protect against:

* Runtime compromise (e.g. injected JS, malicious plugins)
* Memory scraping

---

### Implementation Details

* New IPC handlers:

  * `SAFE_STORAGE_AVAILABLE`
  * `TOKEN_ENCRYPT`
  * `TOKEN_DECRYPT`

* Renderer-side interception:

  * `localStorage` proxy for "token"
  * `document.cookie` proxy for "tokens"

* Main process:

  * Cookie migration logic
  * Encryption via `safeStorage`

---

### Tradeoffs

* Possible forced logout if encryption fails
* Slight IPC overhead for encryption/decryption
* Token still exists in memory during runtime

---

### Result

Removes plaintext token exposure from disk with minimal impact on user experience and without modifying Discord internals.


Polyfish0:
I've done this, because I am a server mod and one of our mods got hacked by this. So I wanted to prevent that easy way of token grabbing. My test code is the following. Executed inside `~/.config/vesktop/sessionData/Local Storage/leveldb`

```python
import plyvel

db = plyvel.DB(".")
for key, value in db.iterator():
    if "_https://discordapp.com\x00\x01token" == key.decode("UTF-8"):
        print(value.decode("utf-8"))
        ```

fckyough:
Why does the copyright note in `tokenProtection.ts` say 2023?

Polyfish0:
I just copied it from https://github.com/Vencord/Vesktop/blob/main/src/renderer/appBadge.ts

Vendicated:
dont send ai generated pull requests.

Polyfish0:
Ah yes, typical incompetence. Anyways, have a great day and continue with getting discord token stolen

Vendicated:
Incompetence is not wanting aislop garbage prs? Sure, man

PR #1251, opened 2026-04-27
fix DTLS Connecting issues for users of tailscale or similar VPNs

### LINKED ISSUE #876: [Bug] strange call behaviour when using VPN
### Discord Account

_No response_

### Operating System

Frankenstein (kinda Arch Linux, but not really)

Wayland

### Package Type

official AppImage, AUR (vesktop and vesktop-bin), running/built from source

When running vesktop with a VPN (in this case WireGuard through Tailscale) joining any call will hang on "DTLS Connecting" each and every time

Joining any call on official discord canary or;
Joining any call w/o VPN;
yields expected result of "call connects completely"

(explained in prior parts, but just in-case)
1. use a VPN (probably related to WireGuard, but don't know how to check w/ any other)

### Debug Logs

potentially related to #756 and #768
"debug" logs include no relevant info, if any electron console logs would help, wei can provide those
checked that it is not DNS setup by restricting `*.discordapp.com` `*.discord.media` and `*.discord.gg` to `9.9.9.9` and then tried `1.1.1.1`

A friend suggested that the issue might be related to DAVE, but (at least officially?) it was released not more than a month ago (announcement 10 days ago), and this issue was present before that
looking @ Electron debug tools console, wei were not able to ascertain there are meaningful differences between pure Vesktop vs Vesktop + VPN output, besides that `peerConnectionState` never transitions into `connected`, but wei don't know what subsystem that is logged by

also, [a potentially relevant tailscale bug report](https://github.com/tailscale/tailscale/issues/10396)

### Request Agreement

Zylanx:
This changes the WebRTC IP handling policy to ignore non-default routes.

When running Tailscale, WebRTC can sometimes get confused and bind to that private VPN interface, hanging forever on DTLS Connecting.
This change, from my understanding, makes WebRTC only bind to the default route interface which is likely the intended behaviour anyway, avoiding weird issues with tailscale.

fixes https://github.com/Vencord/Vesktop/issues/876

### DIFF
--- a/src/main/mainWindow.ts
+++ b/src/main/mainWindow.ts
@@ -409,6 +409,9 @@ function createMainWindow() {
 
     win.webContents.setUserAgent(BrowserUserAgent);
 
+    // Fixes voice/video calls hanging at DTLS Connecting when using VPNs like Tailscale
+    win.webContents.setWebRTCIPHandlingPolicy("default_public_and_private_interfaces");
+
     // if the open-url event is fired (in index.ts) while starting up, darwinURL will be set. If not fall back to checking the process args (which Windows and Linux use for URI calling.)
     // win.webContents.session.clearCache().then(() => {
     loadUrl(darwinURL || process.argv.find(arg => arg.startsWith("discord://")));


VozDeOuro:
any news on that?

toast003:
I have been using a patched version of vesktop with this PR for a while now and it works perfectly

ran-dall:
This fixes the main Discord window, but `setWebRTCIPHandlingPolicy` is scoped per `WebContents`, so applying it only in `createMainWindow()` misses other windows.

Vesktop creates Discord popouts separately: [`makeLinksOpenExternally` allows Discord `/popout` windows via `createOrFocusPopup`, then listens for `did-create-window` and calls `setupPopout` for the newly created popout window](https://github.com/Vencord/Vesktop/blob/61ecf66936ed5069b661306781e396edf26b1e01/src/main/utils/makeLinksOpenExternally.ts#L59-L70). [`createOrFocusPopup` returns `allow` with `overrideBrowserWindowOptions`, so Electron creates the popout as its own `BrowserWindow`](https://github.com/Vencord/Vesktop/blob/61ecf66936ed5069b661306781e396edf26b1e01/src/main/utils/popout.ts#L76-L92).

Maybe we could we centralize this with `app.on("web-contents-created", ...)` instead? That would apply the policy to the main window, Discord popouts, and any future `WebContents` Vesktop creates.

Meteoni:
Can confirm that the same issue exists on Windows. Thanks god the patch worked > it's been driving me crazy for the past few days

wociscz:
> I have been using a patched version of vesktop with this PR for a while now and it works perfectly

I second this. Works without any issue when switching on/off multiple different VPNs.

Raedrim:
Please look into merging this, works perfectly, and it's super annoying having to turn off tailscale, join a voice channel, then turning tailscale back on when dealing with SSH

Axenide:
Hi, please merge this. :)

xoconoch:
Is there a specific reason for this to not be merged yet?

Vendicated:
This is a change for all users that could potentially cause issues. It should probably either be a cli flag or opt in setting

2meito:
Per [electronjs documentation](https://www.electronjs.org/docs/latest/api/web-contents#contentssetwebrtciphandlingpolicypolicy),

`default`:
> Exposes user's public and local IPs. This is the default behavior. When this policy is used, WebRTC has the right to enumerate all interfaces and bind them to discover public interfaces.

`default_public_and_private_interfaces`:
> Exposes user's public and local IPs. When this policy is used, WebRTC should only use the default route used by http. This also exposes the associated default private address. Default route is the route chosen by the OS on a multi-homed endpoint.

ideasafines:
A merge or opt-in setting to fix this would be greatly appreciated.

2meito:
Very noobish attempt, but I [added a drop-down setting on top of Zylanx's branch based on the branch picker](https://github.com/Zylanx/Vesktop/commit/da912c05d1413845bfb029d4ce33342769dd0a75). 

I only applied the setting to `createMainWindow` as I'm not familiar enough with electron to propagate to all additional windows.

The setting also doesn't seem to take affect until after vesktop restarts, although I'm not sure if there needs to be a restart required banner (e.g. toggling plugins) since the branch picker also doesn't notify for a restart.

<br/>

Edit:

Made a new PR (#1283) which fixes a really dumb bug, updates the UI to fit under Miscellaneous, and applies the setting to all current and future windows.

coleleavitt:
This is still a persistent issue I can vouch

x0rzavi:
Facing this issue, please merge the fix :D

Zylanx:
@2meito Thanks for doing a bit extra work on this!
I just realised my email client wasn't sending me notifications so I totally forgot about this.
Good work!

Zylanx:
Also I think realistically there should also be an upstream issue made, either in electron or wherever WebRTC is hosted.
It is definitely not correct behaviour.
I'll look into doing so if no one else has soon
Edit: Looks like maybe [there is already](https://issues.chromium.org/issues/474435503) an issue for it on chromium but it has stalled?

Vendicated:
Closing in favour of #1283

Issue #583, opened 2024-05-06
[Bug] Vesktop opens share-picker multiple times


myryk31415:
### Operating System

OpenSUSE Tumbleweed

### Linux Only ~ Desktop Environment

Hyprland

### What happens when the bug or crash occurs?

When pressing screen-sharing multiple popus open in this order (next one opens when selecting something):
1. hyprland-share-picker
2. Discord/Vesktop screenshare popup for quality etc
3. hyprland-share-picker
4. hyprland-share-picker

If I select the same screen 3 times in hyprland-share-picker it streams as expected, otherwise the stream opens after the Discord popup but doesn't load.

### What is the expected behaviour?

Selecting a screen once should probably be enough.

### How do you recreate this bug or crash?

1. open Vesktop
2. join voicecall
3. select screen sharing
4. now the popups open one after another

### Debug Logs

Idk what I should log, I could provide dbus-monitor if that helps.

### Request Agreement

- [X] I have searched the existing issues and found no similar issue
- [X] I am using the latest Vesktop and Vencord versions
- [X] This issue occurs on an official release (not just the AUR or Nix packages)

AAGaming00:
As a temporary workaround, tick "allow a restore"

zbigniewpaczynski:
Same thing happens on Arch with KDE

Vendicated:
issue with your desktop environment / chromium

Different-Name:
Has anyone made an upstream issue for this? I believe it is chromium and not the desktop environment issue, as OBS has no problems

myryk31415:
Thx! ig that should have been done a while ago. Also great username

Different-Name:
Not a bug with chromium or desktop environment

https://issues.chromium.org/issues/354537406#comment13

> This is also expected. You have to use "Allow restore token" to avoid the second portal dialog. This is because you first get a portal dialog to pick a screen/window to share for the preview dialog and once you confirm, you get another portal dialog to pick a screen/window for the web page itself. For that reason we use the restore token to avoid the second portal dialog and show the same content on the web page that was selected for the preview dialog. OBS doesn't do that, because it works differently.

A picker prompt is created by Vesktop for the `SCREENSHARE` menu, to display the preview
The stream settings are confirmed and then only a second dialogue should be created for the stream itself

The third dialogue should not occur (in my opinion at least). I would assume that it is because discord is creating a preview somewhere in the background resulting in the following:

1. Picker for sharescreen preview
[...]

PR #1253, opened 2026-04-29
screenShare: skip duplicate portal chooser on Wayland

coleleavitt:
## Summary

On non-GNOME Wayland compositors (niri, hyprland, sway, labwc) Vesktop currently shows the xdg-desktop-portal chooser dialog **twice** in a row when a user clicks "Share Screen":

1. First chooser: triggered by `desktopCapturer.getSources()` enumerating sources for Vesktop's preview thumbnails
2. Second chooser: triggered by Chromium's `WebRTCPipeWireCapturer.Start()` when `setDisplayMediaRequestHandler`'s callback fires — this opens a fresh portal session independent from the first one

GNOME's portal silently auto-restores the second session via its in-memory token, masking the issue. Other portals show the dialog twice.

## Fix

On Wayland, pass a synthetic `{ id: "screen:0:0", name: "Entire Screen" }` source to the `displayMediaRequestHandler` callback instead of the real Source object from `getSources()`. Chromium's PipeWire capturer ignores Vesktop-side source IDs on Wayland (it always opens its own portal session for the actual stream regardless), so the synthetic ID is harmless — and PipeWire's portal shows its chooser exactly once.

## Test plan

Tested on niri 26.04 + xdg-desktop-portal-gnome 48 + Vesktop 1.6.5 + Electron 40.4.0 + Chromium 144:
- **Before**: clicking "Share Screen" showed two consecutive GTK4 chooser dialogs; user had to pick a source twice.
- **After**: only one chooser dialog appears; stream starts normally; Discord viewers receive the screenshare.

No regression observed on X11 (the `if (isWayland)` branch is unchanged for non-Wayland).

## Refs

- [electron/electron#30652](https://github.com/electron/electron/issues/30652) — *Redundant app-created screen share dialog on Linux Wayland* (open since 2021)
- [#583](https://github.com/Vencord/Vesktop/issues/583) — Vendicated confirmed Chromium opens two sessions intentionally
- Documented community workaround approach: [comment by xander1421 on electron#30652](https://github.com/electron/electron/issues/30652)

### DIFF
--- a/src/main/screenShare.ts
+++ b/src/main/screenShare.ts
@@ -58,7 +58,15 @@ export function registerScreenShareHandler() {
                 if (stream === null) return callback({});
             }
 
-            callback(video ? { video: sources[0] } : {});
+            // Workaround for electron/electron#30652: Chromium's
+            // WebRTCPipeWireCapturer creates a fresh portal session in Start()
+            // that is independent from the one desktopCapturer.getSources just
+            // used. On non-GNOME compositors (niri, hyprland, sway) this
+            // surfaces as TWO portal chooser dialogs. Passing a synthetic
+            // source ID here lets PipeWire's portal show its picker only
+            // once, when the actual stream starts.
+            const placeholder = { id: "screen:0:0", name: "Entire Screen" } as Electron.DesktopCapturerSource;
+            callback(video ? { video: placeholder } : {});
             return;
         }
 


glitchplaysgames714:
why the hell has this not been merged yet???

Vendicated:
cause it's aislop and a very ugly hack that probably doesn't even work. Or if it does work, it only does because of Chromium/Wayland jank and could break at any time

Vendicated:
Also "why the hell has this not been merged yet???" is no way to communicate and I really shouldn't have to tell you that. Please be polite and less demanding/entitled or you will be blocked

PR #1256, opened 2026-05-04
Feat/browser popup external button

### LINKED ISSUE #1097: [Feature Request] Add "open in browser" for "Open Links In App"
### Discord Account

quadrinex

### Motivation

The functionality of oppening links in app is usefull when you just want to see a post or smt like that, tho there are usecases where you need to open it in browser.

### Solution

Add a "open in browser" button at 1 or 2.
Image

### Alternatives

When right clicking a link you have the option.

### Additional context

_No response_

### Request Agreement

- [x] I have searched the existing issues and found no similar issue
- [x] This is not a plugin request
- [x] This is not a Vencord feature request

LucaHDR:
Add button in browser popup to be able to also open the site inside the browser

image

Closes #1097


src/main/utils/makeLinksOpenExternally.ts
```diff
@@ -49,6 +50,9 @@ export function handleExternalUrl(url: string, protocol?: string): { action: "de
 
 export function makeLinksOpenExternally(win: BrowserWindow) {
     win.webContents.setWindowOpenHandler(({ url, frameName, features }) => {
+        
+        initBrowserPopupMenuBar(win, url);
```
quadrinex, line 54: Review written with ChatGPT.

"initBrowserPopupMenuBar(win, url)" runs before the URL is parsed/validated and before this popup is confirmed to be the in-app browser popup case. Since "Menu.setApplicationMenu()" mutates the global application menu, denied/invalid/non-browser popup URLs can still replace the app menu. This should probably only run after the handler has confirmed that the popup is allowed and handled as the browser popup.

src/windowMenus.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+
+export function initMenuBar(win: BrowserWindow) {
+
+    const menuItems = createBaseMenuBar(win);
+
+    const menu = Menu.buildFromTemplate(menuItems.filter(isTruthy));
+
+    Menu.setApplicationMenu(menu);
```
quadrinex, line 123: Review written with ChatGPT.

This seems like the wrong abstraction layer for a popup-specific action. "Menu.setApplicationMenu()" affects the entire app, not just the popup. A popup-local UI button, context menu, or window-specific menu would avoid leaking popup state into the global app menu.

src/windowMenus.ts
```diff
@@ -0,0 +1,140 @@
...
+        {
+            label: "Open in browser",
+            click() {
+                shell.openExternal(url);
+            },
+        })
+
+    const menu = Menu.buildFromTemplate(menuItems.filter(isTruthy));
+
+    Menu.setApplicationMenu(menu);
```
quadrinex, line 139: Review written with ChatGPT.

This sets the application menu globally, but I don't see any matching restoration of the default menu when the popup closes or when the popup URL is no longer relevant. After opening one popup, the app can keep an "Open in browser" item pointing at a stale URL until another menu update or restart. Please restore the default menu when the popup closes, or avoid modifying the global application menu for popup-specific state.

quadrinex (commented):
Review written by ChatGPT.

arandomhooman:
progress?

LucaHDR:
> progress?

If someone review it and then updates the progress of the PR / says what to change I am happily to change or adjust. I wont fix merge conflicts forever without an proper info what to change / do for an merge in main :D

PR #1257, opened 2026-05-06
Only download required Vencord desktop assets

zw5:
### Repro

1. Open Vesktop on macOS.
2. Use Vesktop > Force Update Vencord.
3. Hit a Vencord asset download failure.
4. Relaunch Vesktop.

On my machine, this left `~/Library/Application Support/vesktop/sessionData/vencordFiles` with only `package.json`. Startup logs then retried `vencordDesktop*.map` and `.LEGAL.txt` downloads, and Vesktop stayed alive with no main window.

### Change

`downloadVencordFiles()` now maps release assets by exact filename and downloads only the four files `isValidVencordInstall()` checks. It also throws a named missing-asset error if the release lacks one of those files.

### DIFF
--- a/src/main/utils/vencordLoader.ts
+++ b/src/main/utils/vencordLoader.ts
@@ -48,13 +48,22 @@ export async function downloadVencordFiles() {
     const release = await githubGet("/repos/Vendicated/Vencord/releases/latest");
 
     const { assets }: ReleaseData = await release.json();
+    const assetsByName = new Map(assets.map(asset => [asset.name, asset]));
+    const missingAssets = FILES_TO_DOWNLOAD.filter(name => !assetsByName.has(name));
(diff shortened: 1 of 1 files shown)


zw5:
Why did you close the issue, are the reproduction steps not clear enough? I actually had to fix this myself because my local vesktop install broke, and not even reinstalling the app using homebrew fixed it. Sure, I’ll give you that the pull request looks like spammy made up LLM issue, but I assure you I verified everything myself

Vendicated:
because it's aislop and stupid

Vendicated:
You're solving a made up issue that does not actually exist with your shitty ai code

Literally peak ai brainrot, your brain is smoosh from all the AI you fed it

Issue #569, opened 2024-05-02
Screenshare broadcasts system-wide audio, not just that of the application. 


TheodorEvald:
**Describe the bug**

When streaming any application using Vesktop, audio from other applications is being picked up and broadcasted.

**Expected behavior**

Vesktop streaming only the audio of the application I selected to screenshare.

**Desktop (please complete the following information):**
 - OS/Distro: Windows 10 LTSC 21H2

**Additional Context**

I tried streaming VLC, Foobar2000, a variety of games and the issue persists.

Vendicated:
yes this is intended

i don't think it's possible to stream just one window with electron api

PolisanTheEasyNick:
Indeed, using electron API we can stream either all sound from Windows system or sound from one tab:
![image](https://github.com/Vencord/Vesktop/assets/39007846/9a57be62-6856-4bbb-ace5-c35fe7bbb192)
For now, we are setting `audio` parameter just to `loopback`:
https://github.com/Vencord/Vesktop/blob/main/src/main/screenShare.ts#L80

Vendicated:
that api has no option to choose a custom window. WebFrameMain refers to an electron frame, aka a frame of your own app. not other apps

PolisanTheEasyNick:
> that api has no option to choose a custom window. WebFrameMain refers to an electron frame, aka a frame of your own app. not other apps

Ohh, I see now. Anyway there's no way to choose another sound source :(

Vendicated:
this issue does not apply to Linux. only Windows

KIRUYXAN:
So it's not possible to fix this?

Communified:
I love this program, but this being an "intended" feature is almost dealbreaking for the app; I can no longer stream in vc with audio without them being able hear all other sources of audio, INCLUDING themselves in the vc; which makes audio completely pointless. If there's truly no fix due to the api, it sounds like the api needs to be different for an issue as major as this; everything else functions as intended and how I'd expect within native discord. In discord, even desktop audio capture does NOT include discord; meaning you hear everything EXCEPT others voices, which is much more acceptable. From what you said does that mean it's impossible to fix even via a plugin strictly due to what the program is built off?

mazeymoos0022:
Ope yeah I didnt see this one, didn't show up when I searched, sorry peeps, it's kinda annoying, and yes I wish there was a fix, but if ya can't fix it, that's all good. Hope there becomes a fix in the future! ❤

[...]

PR #1260, opened 2026-05-09
Added an option to chose camera as a source + an option to use a microphone as audio input

quadrinex:
#569 Solution proposed by @SlyCedix


quadrinex:
Changed it so its only one commit

SlyCedix:
I'm not seeing my camera appear as an option. (I'm on Linux)

OS: Manjaro Linux x86_64
Kernel: Linux 7.0.3-1-MANJARO

Running your fork via pnpm start

The virtual camera is functional, though, and I can use regular camera share to show it (no audio though without a virtual audio device)

e: Just realized that I actually can't stream at all from the vesktop client, only the discord client. Not sure how I didn't notice that.

PR #1262, opened 2026-05-13
Handle discord:// URLs when Vesktop is already running

Slush97:
## Summary

`discord://` invite links are dropped on the floor when Vesktop is already running — the window focuses but no join modal appears. This fixes that for both Linux/Windows (via `second-instance`) and macOS (via post-startup `open-url`).

## Background

#813 added `discord://` URL handling, but only at first launch:

- `createMainWindow()` reads `process.argv` / `darwinURL` and calls `loadUrl()` once.
- The `second-instance` handler in `src/main/index.ts` ignored `cmdLine` entirely (`_cmdLine`).
- The `open-url` handler only stashed the URL into `darwinURL`, which is never read after the initial window creation.

Since most users have Vesktop already running when they click an invite, this is the common path.

## Fix

Both `second-instance` and `open-url` now funnel through a new `handleProtocolUrl()` helper. For `discord://-/invite/CODE` URLs it sends the code via the existing `RPC_INVITE` IPC, which `src/renderer/arrpc.ts` already wires to `InviteActions.resolveInvite` + `INVITE_MODAL_OPEN` — so the join modal pops up without reloading the entire Discord client (preserves voice connection, scroll state, etc.). Non-invite deep links fall back to `loadUrl()` to match the existing first-launch behavior.

## Test plan

- [x] Start Vesktop, then from another terminal: `xdg-open 'discord://-/invite/YVbdG2ZRG4'` — join modal opens, no client reload
- [ ] Same on Windows (clicking an invite link in a browser while Vesktop is open)
- [ ] Same on macOS
- [ ] Cold start (Vesktop closed) still works as before — `loadUrl` path unchanged
- [x] `pnpm testTypes` and `pnpm lint` pass

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -11,10 +11,12 @@ import "./userAssets";
 import "./vesktopProtocol";
 
 import { app, BrowserWindow, nativeTheme } from "electron";
+import { IpcCommands } from "shared/IpcEvents";
 
 import { DATA_DIR } from "./constants";
 import { createFirstLaunchTour } from "./firstLaunch";
-import { createWindows, mainWin } from "./mainWindow";
+import { sendRendererCommand } from "./ipcCommands";
+import { createWindows, loadUrl, mainWin } from "./mainWindow";
 import { registerMediaPermissionsHandler } from "./mediaPermissions";
 import { registerScreenShareHandler } from "./screenShare";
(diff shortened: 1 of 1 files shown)


Vendicated:
shush little aislopper

PR #1263, opened 2026-05-16
include child process audio nodes when sharing app sound

dowoge:
When sharing audio from an app whose audio plays through a child process (e.g. Rhythia's FMOD subprocess), only the directly selected node was linked, so the actual audio never made it into the stream. Walk PPid via /proc/<pid>/status to group sibling audio nodes under their topmost audio-owning ancestor and expand include/exclude filters to cover the whole group. Skipped when granular selection is on.


Curve:
I think it's better to just turn on Granular Select and select the nodes based on their process

PR #1274, opened 2026-06-13
index.ts: add repair flag to CLI

3a74h:
Some window managers don't have a tray, and it can be difficult to get a tray. So I added the `--repair` flag so it doesn't need a tray icon. :3


Vendicated:
Thanks, but please have a look at how other cli arguments are implemented and do it like that instead

https://github.com/Vencord/Vesktop/blob/main/src/main/cli.ts

3a74h:
i got mogged will fix

3a74h:
i think i implemented it right?

Vendicated:
This doesn't really work as intended because it will also start up the app and then abruptly quit. Vesktop has to be refactored a little to properly allow something like this, which I'll probably just do myself

Vendicated:
I ended up doing this via https://github.com/Vencord/Vesktop/commit/4d2fc95b69b6794744674f91a77fc4d50edbe97e

Thanks for your contribution though ❤️

PR #1277, opened 2026-06-28
fix(autoStart): persist Start With System toggle on Windows

jacobrluttrull:
isEnabled() now also checks executableWillLaunchAtLogin which will fix the "Start With System" toggle in Vesktop settings.


Vendicated:
you know you could have just submitted a 1 line change with a 1 line description you didn't have to include all the ai crap

jacobrluttrull:
That is my fault, i honestly just forgot to remove that crap. This was my first time contributing to open-source and was a little nervous i guess. I made a new commit to update it so its not ai slop and updated the comment above as well. Sorry about that Vendicated.

CrimsonFork:
@Vendicated, given that the "AI" usage seems to be a thing you disagree with, you might want to take the associated commit out of the history entirely and do an equivalent rewrite on your own, otherwise @claude is listed as a contributor, which can create the false impression that the project accepts/supports its usage, especially for those who either can't be bothered to look into the specific commits or don't know to, but just go off of the banner you get when you block a user.

Bringing this up because We help maintain a project for tracking slop infested software and got a request to remove Vesktop as a non-slop alternative based off of said banner alone, which We don't blame the user in question for, but they're likely not the only one to see it first thing and write off the project entirely. https://codeberg.org/ai-alternatives/llm-afflicted-software/issues/48

PR #1280, opened 2026-07-06
feat: wayland idle detection native module

### LINKED ISSUE #396: Do not auto idle if user input is active

T1mbits:
A cleaned up #1198 (sorry if this is disliked practice, I ended up annihilating the old commit tree when I was picking through the commits so I just made a new branch). Fixes #396 

Adds a native node module package that enables idle and suspend detection on most wayland linux compositors, specifically those implementing the `ext-idle-notify-v1` protocol. To be used by the [WaylandIdle](https://github.com/Vendicated/Vencord/pull/4375) plugin.

The package has a Dockerfile and build scripts to easily rebuild the prebuilt modules and generate local builds just like libvesktop. See the old PR for why I didn't just extend libvesktop itself (TL;DR I can't program in C++ and nobody who could was available). The notifier is always running in wayland environments as a side effect of the plugin detecting if the native module works, but refreshes don't create additional instances of the notifier as far as I'm aware.

Fully tested on NixOS 26.11 unstable with niri 26.04, compatible with CustomIdle. I don't have another machine that isn't the exact same as the environment I used so testing to make sure this doesn't break something on other platforms would be really appreciated!

I might be tripping but I'm pretty sure on Xorg and every other platform idling works like usual, please tell me if it doesn't so I can tack on an extension that allows this to work on other platforms too


nixRidge:
I'm not sure if it's the plugin or this pull request specifically, but I can't get this to work on Plasma 6.7.3
Sorry I can't help with more info, but this is what the log spits out when launching Vesktop with the Vencord plugin also loaded:

```
APPIMAGE env is not defined, current application is not an AppImage
Vesktop v1.6.5
Enabled Chromium features: GlobalShortcutsPortal
Disabled Chromium features: WinRetrieveSuggestionsOnlyOnDemand, HardwareMediaKeyHandling, MediaSessionService, WaylandWpColorManagerV1
failed to load wayland idle notifier:  Error: /tmp/.org.chromium.Chromium.yF3yjr: file too short
    at process.func [as dlopen] (node:electron/js2c/node_init:2:2625)
    at Object.<anonymous> (node:internal/modules/cjs/loader:2022:18)
    at Object.func [as .node] (node:electron/js2c/node_init:2:2852)
    at Module.load (node:internal/modules/cjs/loader:1584:32)
    at node:internal/modules/cjs/loader:1386:12
    at Module._load (node:electron/js2c/node_init:2:18095)
    at wrapModuleLoad (node:internal/modules/cjs/loader:262:19)
    at Module.require (node:internal/modules/cjs/loader:1607:12)
    at require (node:internal/modules/helpers:153:16)
    at JO (VesktopMain:88:1808) {
  code: 'ERR_DLOPEN_FAILED'
}
Ignore this message if you are on a compositor that doesn't implement ext-idle-notify-v1
[arRPC > ipc] listening at /run/user/1000/discord-ipc-0
[arRPC > websocket] listening on 6463
[arRPC > process] started
```

T1mbits:
Odd, that would be this pull request. I’ll take a look later today or tomorrow but the “file too short” error might mean the .node file is corrupted or not present. If you have docker you can try rebuilding the module locally and see if that helps

nixRidge:
Ah that did the trick. Before rebuilding the module, the files appeared in /tmp but were completely empty. Now they're working. Thanks! :)

T1mbits:
Not sure if this is going to get looked at but this has been working on my machine without problems for the last month

TheStachelfisch:
I hope this gets looked at, it's an important feature that is still missing on Wayland Discord.
It's especially annoying when it comes to getting double notifications on your phone while you're using your computer.

RileyIsPurple:
Will this be added soon? 
This is a great quality of life on Vesktop and other forks from Vesktop, as it finally removes the annoying duplicate notifications coming from your phone. Without having to manually disable the automatic idle detection entirely, which would cause no notifications to come to your phone entirely as long as Vesktop is opened.

PR #1281, opened 2026-07-08
Make rounded corners and shadow customizable

luk1337:
Electron 39 update force enabled both of these, let's have them customizable for those that want these off.


Vendicated:
Thanks !

PR #1287, opened 2026-07-20
fix(macOS): enable system audio for screenshare

### LINKED ISSUE #1091: [Bug] No option to enable screenshare audio on Mac
### Discord Account

_No response_

### Operating System

macOS Sonoma

_No response_

### Package Type

DMG

Whenever I try to screenshare- and this doesn't seem to change whether I select a window or my entire screen- not only does Vesktop not pick up my desktop/application audio, but the screenshare audio source dropdown menu that's shown in the promo images on the main github page doesn't appear at all. I also don't have the usual checkbox when i click the gear icon on my running screenshare.

Ideally, there should be an option to enable screenshare audio or to choose the source like in the Vesktop photos.

1. Join a voice channel
2. Open screenshare dialog

### Debug Logs

```shell
No crash log, but swapped to PTB branch and ran with dev console, nothing out of the ordinary or relating to screenshare audio.
```

### Request Agreement

- [x] I have searched the existing issues and found no similar issue
- [x] I am using the latest Vesktop and Vencord versions
- [x] This issue occurs on an official release (not just the AUR or Nix packages)

heykalachik:
## Summary

Enables the existing "Stream With Audio" toggle in the screenshare picker on
macOS, using Chromium's built-in ScreenCaptureKit / Core Audio Taps loopback
support. 
Closes #1091.

## How it works

- `src/main/index.ts`: unconditionally enables the
  `MacSckSystemAudioLoopbackOverride` Chromium feature on darwin. This is the
  only flag needed — Core Audio Taps (macOS 14.2+) is already enabled by
  default in Chromium 150, and ScreenCaptureKit loopback works below
  macOS 15 without any flag. The override is only relevant on macOS 15+,
  where Chromium disables SCK loopback by default due to a permission-prompt
  bug, but it's the documented way to still get audio
- `src/main/screenShare.ts`: adds `supportsMacLoopbackAudio()` (macOS ≥ 12.3,
  ScreenCaptureKit's minimum), and uses it to (a) tell the renderer via the
  `SCREEN_SHARE_PICKER` IPC payload whether to show the audio toggle
  (`canShareAudio`), and (b) request `streams.audio = "loopback"` when the
  user has the checkbox on
- `src/renderer/components/ScreenSharePicker.tsx`: shows the existing
  "Stream With Audio" `FormSwitch` when `canShareAudio` is true instead of
  gating it on `isWindows`. The capability flag comes from main (which knows
  the real macOS point version) rather than the renderer guessing from
  `navigator.platform`
- `package.json`: adds `NSAudioCaptureUsageDescription` to
  `build.mac.extendInfo`. Without it, Core Audio Taps silently produces a
  muted audio track with no error

## macOS version support

| macOS version | Path | Flag needed |
|---|---|---|
| < 12.3 | Not supported (no ScreenCaptureKit) | toggle hidden |
| 12.3 – 14.1 | ScreenCaptureKit | none (default) |
| 14.2 – 14.x | Core Audio Taps (SCK also available) | none (default) |
| 15.0+ | Core Audio Taps primary; SCK via override | `MacSckSystemAudioLoopbackOverride` (Core Audio Taps needs none) |

## Known limitations

- Loopback captures *all* system audio, not per-application, same as the
  existing Windows behavior. Own-voice echo (Discord's own output looping
  back into the capture) is a known possibility; not addressed here, matches
  Windows parity
- The macOS 12.3 lower bound for showing the toggle is derived from
  ScreenCaptureKit's own minimum OS version and Electron's own (not yet
  merged) docs PR ([electron/electron#47493](https://github.com/electron/electron/pull/47493))


heykalachik:
Hey @Vendicated sorry for bothering you but can you please take a look on this PR?

Vendicated:
@grok have a look

heykalachik:
> @grok have a look

Closing PR since there is no interest in macos users problems lol

Vendicated:
the lack of interest stems from the fact that this PR is aislop

PR #1294, opened 2026-07-30
Windows: fix screenshare containing other participants' voices

vocaldll:
Fixes Windows screenshare audio capturing Vesktop itself, which caused viewers to hear their own voices.

Uses Chromium's `loopbackWithoutChrome` device on supported Windows versions, with normal loopback as a fallback for older builds.

Tested on Windows 11.


wauterboi:
I was able to test this and can verify that it does in fact work! This was the only thing stopping me from using Vesktop for the longest time.

MasterPab:
Works phenomenal. Abysmal.

Vendicated:
Thanks! How did you even find this lmao it's not documented anywhere

vocaldll:
> Thanks! How did you even find this lmao it's not documented anywhere

I found it while looking into Electron’s handling of restrictOwnAudio; [electron/electron#52455](https://github.com/electron/electron/pull/52455) had just added support for it and pointed me to Chromium’s internal kLoopbackWithoutChromeId / loopbackWithoutChrome. From there I traced how Chromium uses it and tried passing the same device ID through Vesktop. Thanks for the merge!

ThaUnknown:
this is actually documented, and exposed to web APIs [as above said]!

https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints/restrictOwnAudio

maxxeddd:
Built it on Windows 10, release ver 19045, with the release check removed and it works completely fine. So could you change it to that instead of 20348? @vocaldll

vocaldll:
> Built it on Windows 10, release ver 19045, with the release check removed and it works completely fine. So could you change it to that instead of 20348? @vocaldll

Thanks for reporting this, I confirmed it works on Windows 10 22H2 (build 19045) and opened #1309.

PR #1296, opened 2026-08-09
Add Back/Forward buttons

Davr1:
This PR adds back the missing back/forward (fasttravel) buttons that the desktop client has
image


Davr1:
\*evilishly unfixes the keybinds\*

Issue #1270, opened 2026-05-31
[Feature Request] Allow users to select default audio source when streaming


juizoi:
### Content

Every time I start a stream, vesktop asks me which audio source I'd like to use, I will always want to share my computer's entire audio, would be great to have an option in settings to allow this!

wankelbauer:
Dear vesktop support,
First, i have arch linux + pipewire..
It would be great if there is an option thats says "always use the (name)matching audiosource"
isnt that possible through matching names?

Or isnt it possible to just "remember last audio source"?
Thanks, kind regards!

Curve:
> Dear vesktop support,
> First, i have arch linux + pipewire..
> It would be great if there is an option thats says "always use the (name)matching audiosource"
> isnt that possible through matching names?
> 
> Or isnt it possible to just "remember last audio source"?
> Thanks, kind regards!

No, chromium does not give use enough information on the selected application to share for us to match it

wankelbauer:
Dear curve, i hardly remember discord doesnt has that extra step when starting a stream? I think u just choose the application in windows and it overtakes the audio of that stream without doing a second step?
Thanks kind regards.

Curve:
> Dear curve, i hardly remember discord doesnt has that extra step when starting a stream? I think u just choose the application in windows and it overtakes the audio of that stream without doing a second step?
> Thanks kind regards. 

Exactly, but this client is different. When doing this:

> u just choose the application in windows

The information exposed to us is too little to match the audio. Chromium/Electron uses opaque handles here that are just IDs that we cant really do anything with.

riukii:
Is it not possible to have an option that remembers the "Entire System" audio selection tho?

Curve:
> Is it not possible to have an option that remembers the "Entire System" audio selection tho? 

That is of course possible

riukii:
If it’s possible, may I ask why it hasn’t been implemented yet?

Curve:
> If it’s possible, may I ask why it hasn’t been implemented yet?

1. No one bothered to do it yet
2. Because every time someone replies like this we feel immense joy by not implementing it :)

juizoi:
> > If it’s possible, may I ask why it hasn’t been implemented yet?
> 
> 1. No one bothered to do it yet
> 2. Because every time someone replies like this we feel immense joy by not implementing it :)

[...]

PR #1310, opened 2026-08-24
Add default audio source setting for linux screen sharing

abishek-sundar:
## Describe your Changes
- Implements [Allow users to select default audio source when streaming](https://github.com/Vencord/Vesktop/issues/1270) 
     - This implementation is only for linux users as "Entire System" uses [venmic](https://github.com/Vencord/Vesktop/blob/cb9c55fe1aa4cd1d3fbe23bddfdca02627c8c372/src/preload/VesktopNative.ts#L87).
- Added a boolean to the Audio Settings modal to default to entire system audio for screen sharing
- Read the above setting in the Screen Share Picker modal 
## Screenshots (if applicable)
- Manually verified that:
     -  when the `defaultToEntireSystem` toggle is enabled, "Entire System" audio source is automatically chosen.
     - when the `defaultToEntireSystem` toggle is disabled, "None" audio source is automatically chosen.
 -  Also ensured settings and behavior persist through application quit/startup. 
image

## Checklist before submitting

### DIFF
--- a/src/renderer/components/ScreenSharePicker.tsx
+++ b/src/renderer/components/ScreenSharePicker.tsx
@@ -299,17 +299,6 @@ function AudioSettingsModal({
                     value={Settings.audio?.deviceSelect ?? false}
                     disabled={Settings.audio?.ignoreDevices}
                 />
-                <FormSwitch
-                    title="Default to Entire System Audio"
-                    description={
-                        <>
-                            Default to <b>Entire System</b> as audio source in the Screen Share Picker.
-                        </>
-                    }
-                    hideBorder
-                    onChange={v => (Settings.audio = { ...Settings.audio, defaultToEntireSystem: v })}
-                    value={Settings.audio?.defaultToEntireSystem ?? false}
-                />
             </div>
         </Modal>
     );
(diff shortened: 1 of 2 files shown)


Curve:
I think it would make sense to just save the last selected option and fallback to entire system if the last option does not exist currently

abishek-sundar:
> I think it would make sense to just save the last selected option and fallback to entire system if the last option does not exist currently

Yeah, agreed. I took a stab at this. I figured it made sense so that once the fallback to "Entire System" triggered, Entire System" should then persist as `lastSelectedSource`, but let me know if you disagree.

Curve:
> > I think it would make sense to just save the last selected option and fallback to entire system if the last option does not exist currently
> 
> Yeah, agreed. I took a stab at this. I figured it made sense so that once the fallback to "Entire System" triggered, Entire System" should then persist as `lastSelectedSource`, but let me know if you disagree. 

Sounds good to me

IToastPotatoes:
this is exactly what i wanted!!!

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -573,10 +573,28 @@ function AudioSourcePickerLinux({
...
+        setIncludeSources(s);
+    };
+
     const [audioSourcesSignal, refreshAudioSources] = useForceUpdater(true);
     const [sources, _, loading] = useAwaiter(() => VesktopNative.virtmic.list(), {
         fallbackValue: { ok: true, targets: [], hasPipewirePulse: true },
-        deps: [audioSourcesSignal]
+        deps: [audioSourcesSignal],
+        onSuccess(value) {
+            if (!value.ok || !Array.isArray(includeSources)) {
```
Curve, line 586: Why the `! Array.isArray` check?

src/renderer/components/ScreenSharePicker.tsx
```diff
@@ -573,10 +573,28 @@ function AudioSourcePickerLinux({
...
+                return;
+            }
+
+            const isAvailable = includeSources.every(selected =>
```
Curve, line 590: Maybe check for `some` here? If one selected e.g. 5 nodes and only one is unavailable it would make sense to still keep the old 4 selected imo

juizoi:
excited to see this PR approved! Been looking forward to this feature for a bit now

PR #1290, opened 2026-07-21
globalshortcuts ptt

Covkie:

PR #1324, opened 2026-09-05
Fix global shortcuts silently degrading to focus bound on relaunch

pendo324:
## Summary

Global shortcuts registered through Electron's `globalShortcut` module on Linux/Wayland+KDE only work while Vesktop has focus after the first relaunch, requiring the user to manually remove and re-add the shortcut in System Settings each time.

## Root cause

Traced via D-Bus tracing (`busctl monitor`) and Chromium source to a real upstream bug in [`GlobalAcceleratorListenerLinux::OnListShortcuts()`](https://github.com/chromium/chromium/blob/65444518f8b2c488969978a14c039ae2defaea3a/ui/base/accelerators/global_accelerator_listener/global_accelerator_listener_linux.cc#L272-L295):

```cc
// If any bound command is not found among the registered shortcuts, bind
// them.
for (const auto& [modified_id, bound_cmd] : bound_commands_) {
  if (!registered_ids.contains(modified_id)) {
    ...
    return;
  }
}
// Every command is already registered with the portal, so no bind pass is
// needed. Mark the session as bound...
bind_state_ = BindState::kBound;
```

Shortcut IDs are derived from a hash of the accelerator group/profile ID (see `GetShortcutPrefix()` in the same file), so they're stable across relaunches. On relaunch, `ListShortcuts` returns the IDs from the *previous* session (the portal remembers them), so every `modified_id` is already `contains`ed, the loop never enters its body, and `BindShortcuts()` is never called for the new session. The shortcut then quietly degrades to focus-bound, while KDE's System Settings still shows it as "registered" (stale, from the old session). This can't be worked around from application code through Electron's `globalShortcut` API, since it's a bug in Electron/Chromium's own binding logic.

Related, though not identical: [electron/electron#52218](https://github.com/electron/electron/issues/52218) documents the adjacent problem that `globalShortcut.register()`/`isRegistered()` can't distinguish bound/pending/denied on this same Wayland portal path at all, since nothing surfaces the `BindShortcuts` response to callers.

## Fix

Replaced Electron's `globalShortcut` on Linux/Wayland with a direct `org.freedesktop.portal.GlobalShortcuts` binding, implemented in a new Rust `GlobalShortcuts` class (via `ashpd`) inside `packages/libvesktop`. It sidesteps the bug entirely by never consulting `ListShortcuts` to decide whether to bind. Instead, every launch unconditionally calls `BindShortcuts` for the current session using a stable, app-generated shortcut ID (the keybind's own UUID from settings), which KDE's `kglobalaccel` recognizes and reuses the existing consent grant for, so there's no repeat prompt.

Verified via D-Bus tracing across multiple consecutive relaunches. Shortcuts bind silently with no re-prompt, and fire correctly from an unfocused window consistently.

## Why a native binding instead of Electron's API

This trades being at the mercy of Electron/Chromium's shortcut implementation (and whatever bugs land in it) for direct control over the actual portal protocol calls. We talk to `xdg-desktop-portal` ourselves, so when something like this bug shows up we can fix it immediately rather than waiting on and tracking an upstream Electron fix. It also opens the door to functionality Electron's `globalShortcut` doesn't expose at all, like real keyup/keydown events (needed for correct push-to-talk as seen in #1290) instead of just a single fire-on-keydown callback.

Unlike at the time when PR #326 was proposed, now Vesktop already has a native module dependency in tree. This PR simply adds (relatively minimal) new functionality to the module, and ports it to Rust (which arguably has a more ergonomic NAPI developer experience).

## What changed

- `packages/libvesktop` ported from C++/node-addon-api to Rust/napi-rs 3
  - `getAccentColor`, `updateUnityLauncherCount`, `requestBackground` are now async (deliberate breaking change)
  - New `GlobalShortcuts` class, exposing a discriminated-union `ShortcutsEvent` payload over a threadsafe-function callback
- Cross-compilation switched from Docker+node-gyp to `cargo zigbuild`, targeting an explicit glibc 2.17 floor for both x86_64/aarch64 without needing a container (verified via `objdump` that the built `.so`'s max required glibc symbol is exactly 2.17)
  - `Dockerfile` kept only as an optional, documented local fallback for a hermetic build (consider replacing `prebuilds/*.node` binaries with CI/Docker built binaries). Uses `fedora:44` since Fedora packages `zig`, `rustup`, and a current `nodejs` natively (Debian needs NodeSource for a current Node and has no `zig` package at all)
- `src/main/globalShortcuts/nativeGlobalShortcuts.ts` lazy-loads the native module, falls back cleanly to Electron's `globalShortcut` if unavailable
- `src/main/globalShortcuts/acceleratorToXdgTrigger.ts` converts Electron accelerator strings to XDG shortcuts spec trigger format
- `src/main/keyBinds.ts` gated on `shouldUseNativeGlobalShortcuts()` (Linux+Wayland only)

## Prior art

- [`tuxinal/venbind`](https://github.com/tuxinal/venbind) and #326.
- [`Milkshiift/goofbind`](https://github.com/Milkshiift/goofbind), a Rust/`ashpd` Wayland global-shortcuts implementation descended from venbind, which our `GlobalShortcuts` class's portal-call sequencing (`CreateSession` → `BindShortcuts` → `Activated`/`Deactivated` streams) is modeled on.
- #1290, a C++ `XDPGlobalShortcuts` prototype (raw GDBus), an earlier attempt at the same native-portal approach this PR takes.


Vendicated:
5000 lines of aislop, incredible

next time save your tokens

PR #1330, opened 2026-09-12
feat: add status switcher to tray menu

medrinnn:
Adds a Status submenu to the tray icon with Online, Idle, Do Not Disturb, and Invisible options.

## Describe your Changes

so i wanna say first of if all it my first ever contribution (so it might be suck)
and im a good js not ts dev soooo i did some googling and ai to search stuff

ok this feature has been copy pasted from fluxer and i added into vesktop
basically it allow you to change status for your tray icon
so the status changing code is taken from [here](https://github.com/nicola02nb/autoSwitchStatus/blob/main/index.tsx#L101l)
if the ipc is not good that bc ai suck at explaining stuff 
im here to take any advice

## Checklist before submitting


medrinnnn:
please close this pr bc i lost access to this github acc if you can close that would be appreciated

PR #1334, opened 2026-09-16
add inactive account notifications

rosset-nocpes:
## Changes

Added an opt-in setting for notifications from inactive Discord accounts, keeping those accounts connected through isolated hidden windows. Notifications will be prefixed with the account name, and clicking one will switch to the correct account and open the exact message. It's an improvement for existing discord account switcher because imho this feature is useless if it only can switch accounts

Each inactive account uses additional memory, so the feature is disabled by default.

## Screenshots

N/A - notification appearance depends on the OS.

## Checklist

- [x] Tested inactive account notifications.
- [x] Tested notification click, account switching, and message navigation.
- [x] Tests pass.


snp5164:
unused ram is wasted ram but only to a certain point bro

Vendicated:
Sorry but this is insane

